PR #4, opened 2020-09-23
CUDA implementation of ANISymmetryFunctions

peastman:
I have the computation of the symmetry functions working.  I still need to do the derivatives.  Then I'll need to create some good test cases for profiling and optimizing.


peastman:
The CUDA implementation is now complete.  Next step is to write a benchmarking program so I can profile it.

PR #9, opened 2020-10-06
Fix a memory leak in CudaANISymmetryFunctions

raimis:
`CudaANISymmetryFunctions::neighborCount` is allocated in the constructor, but nowhere deallocated.


peastman:
Thanks!

PR #18, opened 2020-11-06
CFConv layer for SchNet

peastman:
I've written the CPU implementation so far.  The calculation matches the behavior of SchNetPack, which is different from what's described in the paper.  It's less flexible than SchNetPack though.  For example, it always uses a cutoff, and always uses a cosine cutoff function, where SchNetPack lets you choose an arbitrary cutoff function.

The way I've implemented the derivatives is less than ideal.  The standard method for doing backpropagation assumes you've saved intermediate quantities during the forward pass.  But I don't want to save any per-interaction quantities, so instead I have to repeat nearly all the calculations from the forward pass in the backward pass.  This slows it down a bit, but I think it will still end up being much faster on the GPU than if we save and reload per-interaction data.  Once the CUDA implementation is written we can investigate that assumption and see if it's true.


peastman:
This is ready to review.  I posted benchmark results at https://github.com/peastman/NNPOps/issues/14#issuecomment-727291229.

schnet/CFConv.h
```diff
@@ -0,0 +1,201 @@
...
+     * @param numGaussians   the number of Gaussian basis functions to use, uniformly spaced between 0 and cutoff
+     * @param cutoff         the cutoff distance
+     * @param periodic       whether to apply periodic boundary conditions
+     * @param gaussianWidth  the width of the Gaussian basis functions
+     * @param w1             an array of shape [numGaussians][width] containing the weights of the first dense layer
+     * @param b1             an array of length [width] containing the biases of the first dense layer
+     * @param w2             an array of shape [width][width] containing the weights of the second dense layer
+     * @param b2             an array of length [width] containing the biases of the second dense layer
+     */
+    CFConv(int numAtoms, int width, int numGaussians, float cutoff, bool periodic, float gaussianWidth) :
```
raimis, line 125: `w1`, `b1`,  etc. is described above, but they are missing here.

peastman: Good catch, I should remove those.  The subclass constructors take those arguments, but they don't pass them on to the parent class.

Issue #21, opened 2020-11-19
CGSchNet support


jchodera:
@peastman: Would love to see if we could support the CGSchNet model described in [this excellent paper](https://aip.scitation.org/doi/10.1063/5.0026133) from @brookehus, since this could allow us to support much larger coarse-grained models as part of our ML integration.

peastman:
Is there really any difference from SchNet?  The cfconv layer seems to be nearly identical.  The only difference they mention is that they change the activation function from softplus to tanh, and they don't give any explanation for why they made that change.  (It also seems like a doubtful choice, since bounded activation functions often don't work as well as unbounded ones for hidden layers.)

jchodera:
I'm not quite sure!

BTW, we've recently noticed that ANI uses CELU, which is not C2-continuous. This causes significant problems with some optimizers, and in principle shouldn't be used with MD. They're now retraining with softplus right now. But we should double-check the activation functions we implement are all C2-continuous.

peastman:
Agreed!

brookehus:
- there's no difference from SchNet in terms of the structure! the cgnet code just allows for the activation function to be changed easily in contrast w/ (e.g.) schnetpack (and also has other modularities like writing your own normalization scheme; in general i support a software that easily allows for swapping out these kinds of things, but since peter's influenced a lot of my coding i doubt this is news)
- @peastman re: doubtful choices, i believe we did obtain lower losses with shifted softplus, which is what canonical schnet uses, and what one would expect, as you point out. however, the simulations resulting from models trained with SSP were problematic (very rugged, iirc), especially as system size increases. we looked into it, and this seems to have to do with whether the activation fxn saturates or not. perhaps stronger regularization or model averaging could offset this instead of or in addition to the activation fxn switch. @nec4, feel free to add your knowledge here

nec4:
@peastman : @brookehus gave a good summary. For our systems and CG mapping choices, we found that using Tanh() as an activation function in place of (shifted) Softplus() resulted in more stable simulations using trained models. Of course, there is no problem for users to try any activation function that they wish.

peastman:
Ok, I can add the option to use tanh instead.  Can you look at the API in #18 and see if otherwise it looks good for your purposes?  Since it's based on handwritten CUDA kernels, it obviously won't be as flexible as pure PyTorch code.

brookehus:
@nec4 do you mind taking a look when you have time? let me know if you need anything

nec4:
[...]

PR #23, opened 2020-12-09
Added option to use tanh activation

peastman:
Addresses the requirements of #21.  @brookehus and @nec4 can you confirm this meets your needs?

### DIFF
--- a/schnet/CFConv.h
+++ b/schnet/CFConv.h
@@ -93,7 +93,7 @@ private:
  * 
  * 1. Compute a set of Gaussian basis functions describing the distance between them.
  * 2. Pass them through a dense layer.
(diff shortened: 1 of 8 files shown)


brookehus:
@nec4 can you look at this when you have time?

nec4:
Of course - taking a look now!

nec4 (approved):
This looks good to me! - though I have not run the tests on my local machine. I  have only double checked the forward pass and derivative calculation (the rest is just modifying input arguments and adding a duplicate test using Tanh instead of SSP if I understand correctly). If you would like me to run the tests, I can setup an environment on my end.

peastman:
Thanks!  I already ran the tests.  Just wanted to make sure this has all the features you need.

PR #33, opened 2021-09-20
Reorganize repo

raimis:
This is preparation for a `conda-forge` package.

- [x] Reorganize the files
- [x] Improve the build system
  - [x] Build the library
  - [x] Build the test executables
  - [x] Run the tests
- [x] Update the documentation


src/schnet/CudaCFConv.cu
```diff
@@ -260,11 +260,11 @@ const float* CudaCFConv::ensureOnDevice(const float* arg, float*& deviceMemory,
     return (const float*) attrib.devicePointer;
 }
 
-__device__ float cutoffFunction(float r, float rc) {
+__device__ float cutoffFunction2(float r, float rc) {
```
peastman, line 263: Why renaming these functions?

raimis: Name collision when linking due to functions with the same names in `CudaANISymmetryFunctions.cu`

raimis: Open to suggestions for a better naming scheme.

peastman: How about adding the `static` keyword so it knows those functions are local to this file and shouldn't be exported?

raimis: Fixed!

raimis:
The minimal reorganisation towards a `conda-forge`  package is done.

@mikemhenry could you check if something more is need for a feedstock?

@peastman could you review the rest?

mikemhenry:
This looks good! It shouldn't be too hard to package since all the dependencies are on conda-forge.

raimis:
@peastman can we merge this?

peastman:
Looks good.  Thanks!

PR #38, opened 2021-11-12
Optimize EnergyShifter of TorchANI

raimis:
This assume that a molecule is constant, so the energy contribution from the atomic self-energies can be pre-computed.

- [x] Implement `TochANIEnergyShifter`
- [x] Test
- [x] Documentation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,10 +30,12 @@ foreach(TEST_PATH ${TEST_PATHS})
 endforeach()
 
 add_test(TestBatchedNN         pytest -v ${CMAKE_SOURCE_DIR}/src/pytorch/TestBatchedNN.py)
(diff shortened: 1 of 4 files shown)


raimis:
@peastman could you review?

peastman:
What's the reason for including this here?  It ought to be a trivial operation.  Does TorchANI somehow manage to implement it in an inefficient way?

raimis:
This is based on the assumption that the molecule is constant. TorchANI doesn't support such optimizations.

peastman:
Wouldn't it make more sense to contribute this to TorchANI then?  It seems out of scope for this repository.

raimis:
@peastman can we proceed with this?

peastman:
Yes, go ahead.

PR #40, opened 2021-11-15
PyTorch wrapper for SchNet operations

raimis:
- [x] Wrapper for `CFConvNeighbors`
- [x] Wrapper for `CFConv`
- [x] Tests
- [x] Documentation


raimis:
@peastman this is ready for the review.

src/pytorch/CFConv.cpp
```diff
@@ -0,0 +1,300 @@
...
+        if (input.device() != positions.device())
+            throw std::runtime_error("The device of \"input\" and \"positions\" has to be the same");
+        if (input.scalar_type() != torch::kFloat32)
+            throw std::runtime_error("The type of \"input\" has to be float32");
+        if (input.dim() != 2)
+            throw std::runtime_error("The shape of \"input\" has to have 2 dimensions");
+        if (input.size(0) != positions.size(0))
+            throw std::runtime_error("The size of the 1nd dimension of \"input\" has to be equal to the 1st dimension of \"positions\"");
+
+        if(!impl) {
```
peastman, line 99: I'm not certain if this is guaranteed to work correctly or if you could get uninitialized memory.  I would add `impl(nullptr)` to the constructor initialization list to be sure.  I've been pondering the `shared_ptr` documentation, and it's ambiguous on this.

src/pytorch/CFConv.cpp
```diff
@@ -0,0 +1,300 @@
...
+            throw std::runtime_error("The size of the 1nd dimension of \"input\" has changed");
+        if (input.size(1) != numFilters)
+            throw std::runtime_error("The size of the 2nd dimension of \"input\" has changed");
+        if (input.device() != device)
+            throw std::runtime_error("The device of \"input\" has changed");
+
+        // if (cudaImpl) {
+        //     const torch::cuda::CUDAStream stream = torch::cuda::getCurrentCUDAStream(tensorOptions.device().index());
+        //     cudaImpl->setStream(stream.stream());
+        // }
```
peastman, line 176: You have several blocks that refer to `cudaImpl` and are commented out.  Is this obsolete code that should be removed?  Or is there a reason you've left it in?

src/pytorch/CFConvNeighbors.cpp
```diff
@@ -0,0 +1,97 @@
...
+        throw std::runtime_error(std::string("Encountered error ")+cudaGetErrorName(result)+" at "+__FILE__+":"+std::to_string(__LINE__));\
+    }
+#endif
+
+namespace NNPOps {
+namespace CFConvNeighbors {
+
+using HolderPtr = torch::intrusive_ptr<Holder>;
+
+Holder::Holder(double cutoff) : torch::CustomClassHolder(), cutoff(cutoff), numAtoms(0), device(torch::kCPU), impl() {}
```
peastman, line 46: This should likewise be `impl(nullptr)`.

src/pytorch/CFConvNeighbors.h
```diff
@@ -0,0 +1,52 @@
...
+#include "CFConv.h"
+
+#ifndef NNPOPS_CFCONVNEIGHBORS
+#define NNPOPS_CFCONVNEIGHBORS
+
+namespace NNPOps {
+namespace CFConvNeighbors {
+
+using torch::Device;
+using torch::Tensor;
```
peastman, line 34: The `using` keyword should never be used in a header file.  If you do, it pollutes the namespace of any source file that includes it.

src/pytorch/TestCFConvNeighbors.py
```diff
@@ -0,0 +1,82 @@
...
+        # TODO test the result
+
+    with tempfile.NamedTemporaryFile() as fd:
+
+        torch.jit.script(neighbors_ref).save(fd.name)
+        neighbors = torch.jit.load(fd.name)
+
+        for _ in range(3):
+            neighbors.build(positions)
+            # TODO test the result
```
peastman, line 82: Don't forget your TODOs!  Probably just check that both objects produced the same neighbors?  Or you could check them against the actual distances.

peastman:
This looks really good!  I added just a few minor comments.

src/pytorch/TestCFConvNeighbors.py
raimis: Removed these comments. Nothing more to check here except for crashes.

src/pytorch/CFConvNeighbors.h
raimis: These are within a namespace, so the mapping is `torch::Device` --> `NNPOps::CFConvNeighbors::Device`, not just `Device`. Also, this header is internal and would be needed for the public API.

src/pytorch/CFConv.cpp
raimis: `std::shared_prt()` and `std::shared_ptr(nullptr)` are equivalent (https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr).

src/pytorch/CFConv.cpp
raimis: These are to support the non-default CUDA streams (like https://github.com/openmm/NNPOps/pull/37). I'll finish in another PR and it needs more changes.

src/pytorch/CFConvNeighbors.cpp
raimis: std::shared_prt() and std::shared_ptr(nullptr) are equivalent (https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr).

raimis: OK! Initializing explicitly (https://github.com/openmm/NNPOps/pull/40/commits/15711e2d91fa550fcfac028323d693a109f87cf1)

src/pytorch/CFConvNeighbors.h
peastman: I still wouldn't do it.  This is just one of those rules of C++ development that it's best to follow consistently.  There's little to be gained from breaking it and a lot to lose.  Once you write a header file, you never know where someone else might include that file in the future or what assumptions they'll have made.

raimis: Removed that stuff

peastman:
Looks good to merge whenever it's ready.

Issue #44, opened 2021-12-10
Removing the dependency of TorchANI


raimis:
NNPOps depends on TorchANI but, as discussed in #39, it shouldn't.

@peastman proposal (https://github.com/openmm/NNPOps/pull/39#issuecomment-977020533):
> The first thing I would do is open one or more issues on the TorchANI repo suggesting these changes and offering to provide code. This class doesn't implement any kind of calculation. It's just a workaround for a particular flaw in TorchANI (that it repeats the calculation on every step). The more of these changes we can push upstream to them, the better.
>
> For anything they don't want to accept, the next thing I would do is move the imports inside the classes. You should be able to `import NNPOps` even if TorchANI isn't installed. In practice that means we should never `import torchani` at the top level. It should only be imported inside the specific classes or functions that use it.
>
> Finally we should consider whether NNPOps is the best place for all of these, or whether, for example, OpenMM-ML would be a better place.

raimis:
As a short-term fix, I'll move `import torchani` into the classes.

raimis:
On Tuesday, we decided:
- Move the top level import into classes/functions as a short-term fix -- #45
- Refactor the wrapper of `ANISymmetryFunctions`, so it doesn't depend on TorchANI
- Move the code related to TorchANI into https://github.com/openmm/openmm-ml

epretti:
> NNPOps depends on TorchANI but, as discussed in [#39](https://github.com/openmm/NNPOps/pull/39), it shouldn't.
> * Refactor the wrapper of `ANISymmetryFunctions`, so it doesn't depend on TorchANI

Done in #129.

> * Move the code related to TorchANI into https://github.com/openmm/openmm-ml

See https://github.com/openmm/openmm-ml/pull/110: OpenMM-ML now uses the TorchANI implementation rather than going through NNPOps.

Closing this as I believe everything here is resolved.

PR #45, opened 2021-12-10
Move the TorchANI imports from the top level

raimis:
Part of #44

- [x] `SymmetryFunctions`
- [x] `BatchedNN`


raimis:
@peastman one more PR for to look at.

PR #49, opened 2022-01-25
Refactor the CUDA error check

raimis:
- [x] Refactor the CUDA error check


raimis:
@peastman could you review?

PR #56, opened 2022-05-12
Fix CI failure

raimis:


mikemhenry:
I think there are issues with  this torchani build for some reason: `cuda102py38h7dc3a01_5` You might want to try doing build 4 and see if that works for cuda 10.2. Also I'm not sure how long we should try and support cuda 10.2

raimis:
@peastman could you review?

.github/workflows/ci.yml
```diff
@@ -5,32 +5,40 @@ on:
...
-            cuda: 10.2.89
-            gcc: 8.5.0
-            nvcc: 10.2
-            python: 3.8
-            pytorch: 1.8.0
+            cuda: "10.2.89"
+            gcc: "8.5.0"
+            nvcc: "10.2"
+            python: "3.7"
+            pytorch: "1.11.0"
```
peastman, line 29: Does this mean we're no longer trying to support any version before 1.11?

raimis: Yes, only PyTorch 1.11. `conda-forge` has pinned the version of PyTorch (https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2812), so there will be no more packages for the older versions.

peastman:
Looks good.

PR #57, opened 2022-05-12
[DO NOT MERGE] Test if we can build with pytorch 1.11

mikemhenry:
Running into this error when building with pytorch 1.11 on cuda 10, going to see if first we can build with 1.11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
             gcc: 10.3.0
             nvcc: 11.2
             python: 3.9
-            pytorch: 1.10.0
+            pytorch: 1.11
           # Without CUDA
           - enable_cuda: false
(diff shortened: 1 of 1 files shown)


mikemhenry:
@raimis 

The combination of GCC 7.5.0 + CUDA 10.2 + pytorch 1.11 gives me this error: 

```
CMake Error in CMakeLists.txt:
  Target "NNPOpsPyTorch" requires the language dialect "CUDA17" (with
  compiler extensions).  But the current compiler "NVIDIA" does not support
  this, or CMake does not know the flags to enable it.
```

What confuses me is pytorch 1.10 builds just fine, but I don't see anything in the pytorch 1.11 release notes that make it seem like this would be a problem. 

https://github.com/conda-forge/nnpops-feedstock/pull/3

PR #59, opened 2022-06-07
Message passing operation

raimis:
- [x] Move the code from TorchMD-NET (https://github.com/openmm/NNPOps/pull/58)
- [x] Integration
- [x] Tests
- [x] Documentation

### DIFF
--- a/src/pytorch/messages/passMessages.py
+++ b/src/pytorch/messages/passMessages.py
@@ -14,7 +14,7 @@ def passMessages(neighbors: Tensor, messages: Tensor, states: Tensor) -> Tensor:
     ----------
     neighbors: `torch.Tensor`
         Atom pair indices. The shape of the tensor is `(2, num_pairs)`.
-        The indices can be `[0, num_atoms-1)` or `-1` (ignored pairs).
+        The indices can be `[0, num_atoms)` or `-1` (ignored pairs).
         See for the documentation of `NNPOps.neighbors.getNeighborPairs` for
         details.
     messages: `torch.Tensor`


raimis:
@peastman could you review?

src/pytorch/messages/passMessages.py
```diff
@@ -0,0 +1,60 @@
...
+    Parameters
+    ----------
+    neighbors: `torch.Tensor`
+        Atom pair indices. The shape of the tensor is `(2, num_pairs)`.
+        The indices can be `[0, num_atoms-1)` or `-1` (ignored pairs).
+        See for the documentation of `NNPOps.neighbors.getNeighborPairs` for
+        details.
+    messages: `torch.Tensor`
+        Atom pair messages. The shape of the tensor is `(num_pairs, num_features)`.
+        For efficient, `num_features` has to be a multiple of 32 and <= 1024.
```
peastman, line 22: Are those limitations really necessary?  It's very common for the number of features not to be a multiple of 32.

src/pytorch/messages/passMessagesCUDA.cu
```diff
@@ -0,0 +1,123 @@
...
+    const Accessor<scalar_t, 2> messages,
+    Accessor<scalar_t, 2> new_states
+) {
+    const int32_t i_neig = blockIdx.x;
+    const int32_t i_dir = blockIdx.y;
+    const int32_t i_atom = neighbors[i_dir][i_neig];
+    if (i_atom < 0) return;
+
+    const int32_t i_feat = threadIdx.x;
+    atomicAdd(&new_states[i_atom][i_feat], messages[i_neig][i_feat]);
```
peastman, line 28: You can eliminate the limitations on number of features by just rewriting this as a loop.

```c++
for (int32_t i_feat = threadIdx.x; i_feat < num_features; i_feat += blockDim.x)
    atomicAdd(&new_states[i_atom][i_feat], messages[i_neig][i_feat]);
```

src/pytorch/messages/passMessages.py
```diff
@@ -0,0 +1,60 @@
...
+    Given a set of `num_atoms` atoms (each atom has a state with `num_features`
+    features) and a set of `num_neighbors` neighbor atom pairs (each pair has a
+    message with `num_features` features), the messages of the pairs are added
+    to the corresponding atom states.
+
+    Parameters
+    ----------
+    neighbors: `torch.Tensor`
+        Atom pair indices. The shape of the tensor is `(2, num_pairs)`.
+        The indices can be `[0, num_atoms-1)` or `-1` (ignored pairs).
```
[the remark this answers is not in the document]
raimis: Thanks! Good catch.

src/pytorch/messages/passMessages.py
raimis: In contrary, the number of internal features is always factor of 32 (e.g. in TorchMD-NET, I have seen usage of 64, 128, 256). GPU computes in warps of 32 threads, so it is the best to match that patter for computational efficiency.

src/pytorch/messages/passMessagesCUDA.cu
raimis: Apart from solving a non-existing problem, this would make the memory access not coalesced and reduce speed.

peastman: It would have no effect on speed at all.  If the number of features happens to satisfy your current requirement, the behavior would be identical to what it currently does.  The `atomicAdd()` would be executed once by every thread with `i_feat` equal to `threadIdx.x`.  The only change would be if the number doesn't satisfy your current requirements, either because it's not a multiple of 32 or it's more than 1024.  In that case it would produce correct behavior, unlike the current code.  So there's no downside at all.

src/pytorch/messages/passMessages.py
peastman: I frequently create models that don't satisfy those requirements, including in TorchMD-Net.  For example, I've trained models with 48 or 80 features per layer.

raimis: Why?

src/pytorch/messages/passMessagesCUDA.cu
raimis: It will reduce the number of thread by the number of features. The reduced parallelism would result into reduced speed.

peastman: I'm not suggesting any change to the number of threads.  The only thing I'm suggesting is wrapping the `atomicAdd()` in a loop as shown above.  If `num_features` happens to match your current restrictions, nothing will change.  Every thread will still call it exactly once.

src/pytorch/messages/passMessages.py
peastman: I don't understand your question.  Why not?  The number of features is a hyperparameter.  It's one of many hyperarameters you tune to balance training accuracy, overfitting, speed, and memory use.  Why place arbitrary limits on it when there's no need to?

PR #61, opened 2022-07-07
`getNeighborPairs` compute the displacement vector

raimis:
- [x] Compute the vector
  - [x] CPU
    - [x] Forward pass
  - [x] CUDA
    - [x] Forward pass
    - [x] Backward pass
- [x] Update tests
- [x] Update documentation

### DIFF
--- a/src/pytorch/neighbors/TestNeighbors.py
+++ b/src/pytorch/neighbors/TestNeighbors.py
@@ -89,6 +89,48 @@ def test_neighbors(device, dtype, num_atoms, cutoff, all_pairs):
     assert np.allclose(ref_deltas, deltas, equal_nan=True)
     assert np.allclose(ref_distances, distances, equal_nan=True)
 
+@pytest.mark.parametrize('dtype', [pt.float32, pt.float64])
+@pytest.mark.parametrize('num_atoms', [1, 2, 3, 4, 5, 10, 100, 1000])
+@pytest.mark.parametrize('grad', ['deltas', 'distances', 'combined'])
+def test_neighbors_grad(dtype, num_atoms, grad):
+
+    if not pt.cuda.is_available():
+        pytest.skip('No GPU')
+
+    # Generate random positions
+    positions = 10 * pt.randn((num_atoms, 3), dtype=dtype)
+
+    # Compute values
+    positions_cpu = positions.detach().cpu()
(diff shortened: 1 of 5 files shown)


raimis:
@peastman can you review?

peastman:
`test_neighbor_grads()` checks that the CPU and CUDA versions give the same results, but it doesn't check that they're actually correct.  How about adding a test for correctness?  For distances and deltas, it's easy to compute analytically what the expected gradients are.

raimis:
@davkovacs I have built a conda package from the latest commit (https://github.com/openmm/NNPOps/pull/61/commits/32eeae51f2064baeb4dccaca7b67dea6f394ec95).
```
conda install -c conda-forge -c raimis/label/pr61 nnpops==0.3rc0
```
Let me know if it works as you wished and I'll finish the PR.

davkovacs:
Would it work with PyTorch 1.12? We want to compile our models with torchscript to create the TorchForce object, but have some operations that cna only be compiled with the latest PyTorch.

raimis:
At the moment, `conda-forge` pins to PyTorch 1.11 (https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/ed6ced1ec023a07826211329f97bcf75466fdc97/recipe/conda_build_config.yaml#L728).

If you want PyTorch 1.12, you have to build from the source.

davkovacs:
I am going on holiday now, so I will not be able to test this thoroughly for the next couple of weeks. But will be one of the first things when I am back.

raimis:
`conda-forge` started to migrate to PyTorch 1.12.

raimis:
> test_neighbor_grads() checks that the CPU and CUDA versions give the same results, but it doesn't check that they're actually correct. How about adding a test for correctness? For distances and deltas, it's easy to compute analytically what the expected gradients are.

For CUDA, I write the backward pass explicitly. For CPU, the backward pass is implicit and relies on the autograd. So, two independent implementation of gradients agree. This should be enough to prove correctness.

peastman (approved):

raimis:
@davkovacs `conda-forge` started to migrate to PyTorch 1.12. There will be new packages, when the dependencies migrated (https://github.com/conda-forge/nnpops-feedstock/issues/9)

PR #70, opened 2022-11-11
getNeighborPairs() supports periodic boundary conditions

peastman:
I've implemented the CPU version but not the CUDA version so far.  Please take a look and see if the API and implementation look OK.

### DIFF
--- a/src/pytorch/neighbors/getNeighborPairsCPU.cpp
+++ b/src/pytorch/neighbors/getNeighborPairsCPU.cpp
@@ -28,7 +28,7 @@ static tuple<Tensor, Tensor, Tensor> forward(const Tensor& positions,
 
     TORCH_CHECK(cutoff.to<double>() > 0, "Expected \"cutoff\" to be positive");
 
-    if (box_vectors.dim() != 0) {
+    if (box_vectors.size(0) != 0) {
         TORCH_CHECK(box_vectors.dim() == 2, "Expected \"box_vectors\" to have two dimensions");
         TORCH_CHECK(box_vectors.size(0) == 3 && box_vectors.size(1) == 3, "Expected \"box_vectors\" to have shape (3, 3)");
     }
@@ -47,7 +47,7 @@ static tuple<Tensor, Tensor, Tensor> forward(const Tensor& positions,
 
     Tensor neighbors = vstack({rows, columns});
     Tensor deltas = index_select(positions, 0, rows) - index_select(positions, 0, columns);
-    if (box_vectors.dim() == 2) {
+    if (box_vectors.size(0) != 0) {
         deltas -= outer(round(deltas.index({Slice(), 2})/box_vectors.index({2, 2})), box_vectors.index({2}));
         deltas -= outer(round(deltas.index({Slice(), 1})/box_vectors.index({1, 1})), box_vectors.index({1}));
         deltas -= outer(round(deltas.index({Slice(), 0})/box_vectors.index({0, 0})), box_vectors.index({0}));
--- a/src/pytorch/neighbors/getNeighborPairsCPU.cpp
+++ b/src/pytorch/neighbors/getNeighborPairsCPU.cpp
@@ -31,6 +31,20 @@ static tuple<Tensor, Tensor, Tensor> forward(const Tensor& positions,
     if (box_vectors.size(0) != 0) {
         TORCH_CHECK(box_vectors.dim() == 2, "Expected \"box_vectors\" to have two dimensions");
         TORCH_CHECK(box_vectors.size(0) == 3 && box_vectors.size(1) == 3, "Expected \"box_vectors\" to have shape (3, 3)");
+        double v[3][3];
+        for (int i = 0; i < 3; i++)
+            for (int j = 0; j < 3; j++)
+                v[i][j] = box_vectors[i][j].item<double>();
+        double c = cutoff.to<double>();
+        TORCH_CHECK(v[0][1] == 0, "Invalid box vectors: box_vectors[0][1] != 0");
+        TORCH_CHECK(v[0][2] == 0, "Invalid box vectors: box_vectors[0][2] != 0");
+        TORCH_CHECK(v[1][2] == 0, "Invalid box vectors: box_vectors[1][2] != 0");
+        TORCH_CHECK(v[0][0] >= 2*c, "Invalid box vectors: box_vectors[0][0] < 2*cutoff");
+        TORCH_CHECK(v[1][1] >= 2*c, "Invalid box vectors: box_vectors[1][1] < 2*cutoff");
+        TORCH_CHECK(v[2][2] >= 2*c, "Invalid box vectors: box_vectors[2][2] < 2*cutoff");
+        TORCH_CHECK(v[0][0] >= 2*v[1][0], "Invalid box vectors: box_vectors[0][0] < 2*box_vectors[1][0]");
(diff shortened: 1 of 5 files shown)


peastman:
I added the CUDA implementation.  It mostly works, but `test_neighbor_grads()` fails with an error I'm not sure of the best way to handle:

> RuntimeError: function torch::autograd::CppNode<Autograd> returned an incorrect number of gradients (expected 4, got 3)

Since I added `box_vectors` as a fourth argument, autograd expects it to return the gradient with respect to that argument.  But we don't calculate it, and I'm not sure it would even really make sense.  Any gradient with respect to box vectors will be full of discontinuities.

raimis:
Just return an empty tensor `Tensor()` to indicated that the argument is not differentiable.

peastman:
Thanks!  I made the change and the test now passes.

When I run the complete `TestNeighbors.py` suite, I still get errors in the CUDA version of `test_periodic_neighbors()`:

> RuntimeError: CUDA error: device-side assert triggered

After a while of debugging, I figured out it doesn't really have anything to do with that test.  It's actually caused by `test_too_many_neighbors()`, which intentionally triggers an assertion.  The error condition somehow isn't getting cleared, such that all CUDA tests run after it always fail.  You can observe this by simply adding the line `pt.cuda.synchronize()` to the end of `test_too_many_neighbors()`.  That will cause it to always fail.

peastman:
Any suggestions about what to do with `test_too_many_neighbors()`?  As far as I can tell `torch.cuda` doesn't provide any way to reset the device.  Once an assert has been triggered, there's no way to clear it and any further CUDA operation in that process will fail.

The obvious solution is not to run that test on CUDA.

raimis:
One option is to call `cudaResetDevice` using `ctypes` (https://docs.python.org/3/library/ctypes.html).

raimis:
Each test is run in a separate process. So, after the device is reset PyTorch will follow with normal initialization for the next test.

peastman:
You're welcome to see if you can figure out a way to get it to work.  But in the mean time, let's not hold up a useful feature over a broken unit test that isn't even related to the new feature.

peastman:
It's all ready for review.

raimis:
Great! I'll look at it.

src/pytorch/neighbors/getNeighborPairsCPU.cpp
```diff
@@ -25,6 +28,11 @@ static tuple<Tensor, Tensor, Tensor> forward(const Tensor& positions,
 
     TORCH_CHECK(cutoff.to<double>() > 0, "Expected \"cutoff\" to be positive");
 
+    if (box_vectors.size(0) != 0) {
```
raimis, line 31: It could check here if all the requirements are satisfied.

src/pytorch/neighbors/getNeighborPairsCUDA.cu
```diff
@@ -100,6 +114,12 @@ public:
         TORCH_CHECK(max_num_neighbors_ > 0 || max_num_neighbors_ == -1,
             "Expected \"max_num_neighbors\" to be positive or equal to -1");
 
+        const bool use_periodic = (box_vectors.size(0) != 0);
+        if (use_periodic) {
```
raimis, line 118: It could check here if all the requirements are satisfied too.

raimis (commented):
I think, all the requirements of `box_vectors` can be checked. It would prevent some invalid simulations. The performance impact would be minimal or none (if the CUDA Graphs are used).

src/pytorch/neighbors/getNeighborPairsCUDA.cu
peastman: Is there any way to check it efficiently?  The shape of the tensor is known by the CPU, but the values of its elements are stored on the GPU.  If we want to throw an exception based on the values of elements, that will require device to host data transfers and add significant latency every time it's called.

raimis: Yes, you are right, it won't be efficient. So the checks have to be done on a GPU.

The only problem is how to abort a kernel elegantly. `assert ` does the jobs, but later a GPU needs a reset. In the CUDA docs, I don't see anything better.

peastman: GPU asserts aren't really effective as a way of catching user errors.  In addition to the fact that you can't recover from them and have to reset the whole device, they don't provide any useful information to the user.  They just get a cryptic "device-side assert triggered" message that tells nothing about what the problem was or how to fix it.  The user will usually conclude that your library is broken.

src/pytorch/neighbors/getNeighborPairsCPU.cpp
peastman: I added the checks in the CPU version.

src/pytorch/neighbors/getNeighborPairsCUDA.cu
raimis: I agree, we need to choose between two evils:
- Users get cryptic messages and need to reset the GPU.
- Users get incorrect results silently.

raimis:
@RaulPPelaez how do you handle kernel errors in your code?

We need something to be:
- Compatible with CUDA graphs
- Low overhead
- Clear error messages for users

RaulPPelaez:
AFAIK there is no clean way to assert with CUDA. As you mentioned device assert leaves the CUDA context in an unusable state.
What I normally end up doing is to have some errorState array/value in device (or managed) memory. A thread in a kernel encountering an error atomically writes to this errorState and returns as fast as possible. Then you delay as much as possible checking this value for errors.
For instance, if you at least have a record of this error state, the user can query it manually with some kind of checkErrorState() when he notices results are incorrect (unless the code just crashes, that is).
I have never found a clean way to do this without requiring some kind of synchronization (like a device-host copy or a stream sync).

RaulPPelaez:
Something like this https://github.com/RaulPPelaez/UAMMD/blob/f0447444ef1f9d7520e661b6d7151c16b9cfdb1f/src/Interactor/NeighbourList/CellList/CellListBase.cuh#L223-L244

RaulPPelaez:
> Any suggestions about what to do with `test_too_many_neighbors()`? As far as I can tell `torch.cuda` doesn't provide any way to reset the device. Once an assert has been triggered, there's no way to clear it and any further CUDA operation in that process will fail.
> 
> The obvious solution is not to run that test on CUDA.

What is the intended way of using this functionality? 
A priori one does not know the total number of pairs, right? I understand it is required, or at least useful, to have control of the maximum number of neighbors per particle from outside, but how does one use it in practice?
In the past I have done things like: set 32 maximum neighbours, if building fails because it is too low increase by 32 until it no longer fails.

If something like that is the case here an extra parameter could be passed to choose whether or not to synchronize and check for a tooManyNeighbours error flag, to find the max number of neighbours as a precomputation. When constructing the CUDA graph this check would be omitted.

peastman:
> What I normally end up doing is to have some errorState array/value in device (or managed) memory.

That sounds like a good approach.

Let's merge this now and add error checking along those lines in a separate PR.  That's going to require significant design to figure out an efficient mechanism for the error reporting.

raimis (approved):

raimis:
@peastman let's merge this!

@RaulPPelaez could you open a dedicated issue to discuss and design the error check?

PR #71, opened 2022-11-28
Workaround for pytorch bug

peastman:
Fixes https://github.com/openmm/openmm/issues/3812.


raimis (approved):
Thanks for tracking this bug.

Could you add a comment with reference to the issue? So nobody simplifies the code with `torch.mean` in the future.

peastman:
Done!

PR #75, opened 2022-12-13
Update the CI versions

raimis:
- [x] Include PyTorch 1.12
- [x] Update from the source installation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,9 @@ jobs:
 
     - name: Install dependencies
       run: mamba env create -n nnpops -f environment.yml
+      env:
+        # Needed to install pytorch-gpu on a machine without a GPU
+        CONDA_OVERRIDE_CUDA: ${{ matrix.nvcc }}
 
     - name: List conda environment
       run: |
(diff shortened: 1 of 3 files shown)


raimis:
@peastman could you review?

.github/workflows/ci.yml
```diff
@@ -15,30 +15,46 @@ defaults:
...
+          - name: Linux (CUDA 11.2, Python 3.9, PyTorch 1.11)
+            enable_cuda: true
+            cuda: "11.2.2"
+            gcc: "10.3.*"
+            nvcc: "11.2"
+            python: "3.9.*"
+            pytorch: "1.11.*"
+
+          # Latest supported versions (with CUDA)
+          - name: Linux (CUDA 11.2, Python 3.10, PyTorch 1.12)
```
peastman, line 44: All of these are older versions.  How about CUDA 11.8, Python 3.11, and PyTorch 1.13?

raimis: Python 3.11 and PyTorch 1.13 are not available because there are no corresponding TorchANI packages.

raimis: Changed to CUDA 11.7. CUDA 11.8 isn't available because of the `cuda-toolkit` action not yet supporting it.

peastman (approved):
Looks good.

PR #80, opened 2023-01-16
Add error checking to CUDA version of getNeighborPairs

RaulPPelaez:
I have added a flag in managed memory. It will be atomically written if too many neighbors are found for some particle.
It is checked using an event just after execution of the kernel.
 
All tests are passing (even the too_many_neighbors one in the GPU)


RaulPPelaez:
The Autograd class is not allowed to hold any state, right?
As of  now, the forward function must create and delete and event every time it is called. Any ideas on how to take event creation elsewhere?

src/pytorch/neighbors/getNeighborPairsCUDA.cu
```diff
@@ -151,6 +166,13 @@ public:
                 get_accessor<scalar_t, 2>(deltas),
                 get_accessor<scalar_t, 1>(distances),
                 get_accessor<scalar_t, 2>(box_vectors));
+	    cudaEventRecord(event, stream);
+	    cudaEventSynchronize(event);
```
raimis, line 170: This may hurt performance. CPU will be waiting for the neighbor kernel to finish rather than working on dispatching the next kernel.

src/pytorch/neighbors/getNeighborPairsCUDA.cu
```diff
@@ -151,6 +166,13 @@ public:
...
+	    //Check the error flag
+	    TORCH_CHECK(tooManyNeighborsErrorFlag == 0, "Some particle has too many neighbours, found " +
```
raimis, line 172: This won't work with CUDA Graphs, because the CPU code will be executed only during a graph creation.

RaulPPelaez:
Can't one add a host node for that?

src/pytorch/neighbors/getNeighborPairsCUDA.cu
RaulPPelaez: This will definitely hurt performance, but it is the only way to get the value from the error flag before this function exits. I put it in managed memory, but a synchronization barrier is needed anyway. I believe an event like this is the less invasive form of sync in this case.
The sync might be delayed until line 158 here:
https://github.com/openmm/NNPOps/blob/8b2d427888e548dbb655b6233060e3e8038e92c2/src/pytorch/neighbors/getNeighborPairsCUDA.cu#L156-L159

If the function must guarantee to throw I cannot think of any other way.

src/pytorch/neighbors/getNeighborPairsCUDA.cu
RaulPPelaez: I am trying to wrap the host-side check inside a cudaLaunchHostFunc callback, which is compatible with cuda graphs. Moreover, doing so eliminates the need of explicit synchronizations, since managed-memory wise these callbacks are equivalent to the event synchronization currently in place.
It seems, however, that throwing an exception from inside a callback is not easy. I tried to implement it, but the python test is not able to capture the exception, which terminates the execution instead. 
Do you have some experience with cudaLaunchHostFunc and exceptions?

peastman:
What about moving the host side checking of the flag into the backward pass?  There will usually be a lot of other kernels launched between the two, so waiting on the event at that point won't hurt performance.  The disadvantage is that the check will be missed if the backward pass is skipped, for example if someone computes only energy but not forces.

raimis:
What if a user runs just the forward pass?

peastman:
Right, that's the disadvantage.  It would give people error checking in most of the common use cases without significantly hurting performance.  But there exist use cases where error checking would be skipped.

RaulPPelaez:
Since this is going to be part of a graph the check should go after launching the graph, and the entity doing so should be the one checking for errors in its execution. What function is going to be building and launching the graph?

An exception from inside a CUDA graph is problematic. One solution is triggering a cudaError during graph execution.
For instance setting cudaErrorInvalidValue by calling cudaMemcpy(NULL, NULL, 0, 0). This would appear the next time the cuda error state is checked after launching the graph.

OTOH we could report the error in the two ways: A direct call to forward can check the host flag and launch and throw an exception. A cudaError can be raised if the call to the forward kernel is happening as part of a CUDA graph execution.

RaulPPelaez:
This commit introduces a new optional bool flag, check_errors, to getNeighborPairs. The default (False) will check for errors and throw asynchronously, printing a meaningful message but crashing.
Passing True will instead synchronize and raise the error synchronously, such that the exception can be handled. The same message is printed in both cases.

In case of a cuda graph the False option is forced, the error is thrown asynchronously, crashing the code with a meaningful message. 

If False is chosen error checking is virtually free. There is no synchronization penalty and since the error flag lives in managed memory there should not be a mem transfer footprint at all if the error did not happen.

raimis:
- **Regular mode** (`check_errors = False`): @RaulPPelaez solution is with `cudaLaunchHostFunc` is very good.
  - Pros
    - A user will get a meaningful error message
    - The overhead is insignificant
    - It is compatible with CUDA Graphs
  - Cons
    - The error will happen asynchronously
    - There is no way to catch the error
- **Advanced mode** (`check_errors = True`): I think, we `getNeighborPairs` should just return the number of detected pairs and it is up to the user to check if there was enough space for space. This eliminates the cons of the regular mode by passing the burden of error checking to the users. The current approach with synchronization is only good for debugging, not production. The same effect can be achieved with the debug mode of PyTorch, where it runs the kernel synchronously.

RaulPPelaez:
> * The same effect can be achieved with the debug mode of PyTorch, where it runs the kernel synchronously.

While this would make the error synchronous, it would be non catchable AFAIK. Do you know if Pytorch defines something that can be detected C++ side when using debug mode? That way I could make the exception synchronous AND catchable in debug mode only.

Also, I believe the meaning of check_errors should be the opposite as you wrote. 
- True should check and throw asynchronously.
-  False would just not even check host-side

In both cases the kernel could write a especial value (say -1 or NaN) to, for instance `neighbors[1][max_pairs-1]` when some particle has too many neighs. At least with the current format and strategy I think this can be implemented without overhead when there is no error.

RaulPPelaez:
I followed @raimis suggestion and added a bit of mine, ending up with the following:
hree behaviors are possible:
  1. Default (both false). Operation is CUDA-graph compatible and an uncatchable exception is thrown in case of number of pairs being too high.
  2. check_errors=True. Operation is CUDA-graph compatible. No exception is thrown and the number of found pairs is returned as a fourth output argument, which can be higher than max_number_pairs.
  3. check_errors=False and sync_exceptions=True. Operation is NOT CUDA-graph compatible. The operation synchronizes to check for errors and throws a catchable exception if necessary.

**Pros**:
 -Default behavior is what is same as without this PR but CUDA-graph compatible.
 -User can choose to make exceptions recoverable.
 -Users can choose to manage errors themselves and keep things CUDA-graph compatible.
**Cons**:
 -There are two parameters dealing with just error reporting. 
 -No unit test can be written to test the default behavior, since it crashes the code without pytest being able to overcome it.

The current unit test for this functionality might help understand how it works:
```python
def test_too_many_neighbors(device, dtype):
    if not pt.cuda.is_available() and device == 'cuda':
        pytest.skip('No GPU')
    # 4 points result into 6 pairs, but there is a storage just for 4.
    positions = pt.zeros((4, 3,), device=device, dtype=dtype)
    with pytest.raises(RuntimeError):
        # checkErrors = False will throw due to exceeding neighbours
        # syncExceptions = True makes  this exception catchable at the
        # expense of performance (even when no error ocurred)
        getNeighborPairs(positions, cutoff=1, max_num_neighbors=1, check_errors=False, sync_exceptions=True)
        pt.cuda.synchronize()

    # checkErrors = True will never throw due to exceeding neighbours,
    # but  will return  the number  of pairs  found.
    # syncExceptions is ignored in this case
    neighbors, deltas, distances, number_found_pairs = getNeighborPairs(positions, cutoff=1, max_num_neighbors=1, check_errors=True)
    assert number_found_pairs == 6
```

RaulPPelaez:
I made some changes to make getNeighborPairs CUDA-graph compatible, now one can do something like:
```python
    device = 'cuda'
    dtype = pt.float32
    num_atoms = 100
    positions = 10 * pt.randn((num_atoms, 3), device=device, dtype=dtype)
    cutoff = 5
    graph = pt.cuda.CUDAGraph()
    with pt.cuda.graph(graph):
        neighbors, deltas, distances = getNeighborPairs(positions, cutoff=cutoff, max_num_neighbors=num_atoms*num_atoms)

    graph.replay()
    pt.cuda.synchronize()
```

src/pytorch/neighbors/getNeighborPairs.py
```diff
@@ -44,7 +49,19 @@ def getNeighborPairs(positions: Tensor, cutoff: float, max_num_neighbors: int =
         The vectors defining the periodic box.  This must have shape `(3, 3)`,
         where `box_vectors[0] = a`, `box_vectors[1] = b`, and `box_vectors[2] = c`.
         If this is omitted, periodic boundary conditions are not applied.
-
+    check_errors: bool, optional
```
raimis, line 52: `check_errors=True` does not check errors, `check_errors=False` checks errors. This counterintuitive. Can you reverse the logic?

src/pytorch/neighbors/getNeighborPairs.py
```diff
@@ -44,7 +49,19 @@ def getNeighborPairs(positions: Tensor, cutoff: float, max_num_neighbors: int =
...
+        If set to True the function does not throw due to a number of pairs larger than the maximum.
```
raimis, line 53: In Python, exceptions are raised not thrown. Also, could you document what exception is raised?

src/pytorch/neighbors/getNeighborPairs.py
```diff
@@ -63,18 +80,27 @@ def getNeighborPairs(positions: Tensor, cutoff: float, max_num_neighbors: int =
         If an atom pair is separated by a larger distance than the cutoff,
         the distance is set to `NaN`.
 
+    number_found_pairs: `Optional[torch.Tensor]`
+        Present  if check_errors=True.  Contains the  total number  of
```
raimis, line 84: For consistency, the number of pairs should be reported always.

RaulPPelaez: This changes the interface beyond the new functionality, invalidating any code that uses this function up until that moment.
Every test fails with:
```bash
FAILED TestNeighbors.py::test_is_cuda_graph_compatible - ValueError: too many values to unpack (expected 3)
```

RaulPPelaez:
BTW performance is looking good for the CUDA graph version!
This is time to call getNeighborPairs for a random cloud of particles with periodic boundary conditions and a density such that the number of neighbors is around 128.

RaulPPelaez:
You may review again @raimis

raimis:
@peastman do you have any comments?

peastman:
Does it really make sense to have separate arguments for `check_errors` and `sync_exceptions`?  Is there ever a situation where it makes sense to pass different values for them?  Error checking doesn't work correctly if `sync_exceptions` is false, and it's completely ignored if error checking is disabled.

src/pytorch/neighbors/getNeighborPairsCUDA.cu
```diff
@@ -152,11 +202,33 @@ public:
...
+	if(checkErrors){
+	  static constexpr cudaHostFn_t h_fn = checkTooManyNeighbors;
+	  static std::tuple<int, bool> h_fn_data;
+	  h_fn_data = {max_num_neighbors_, syncExceptions};
+	  cudaLaunchHostFunc(stream, h_fn, (void*)&h_fn_data);
+	  // With chekErrors=false and syncExceptions=false the state of
+	  // the tooManyErrorsFlag is checked  and exceptions are thrown
+	  // asynchronously and  in a  way compatible with  CUDA graphs.
+	  // However,  this  way  of  throwing  an  exception  makes  it
+	  // uncatchable, crashing  the code.
```
peastman, line 215: We should never ever do anything that causes a crash!

RaulPPelaez:
We decided to change the interface so that the number of pairs is always returned, meaning the user can now easily check if the maximum number of pairs is exceeded. This changes the restrictions of the original problem a bit.
 
We wanted the user to be informed (via an exception, for instance) in the case of the number of pairs found being larger than the maximum allowed.

Alas, informing the user in a recoverable way requires synchronizing (slow and incompatible with CUDA graphs), so I believe it is sensible that this functionality is guarded behind a flag.

Best we can do AFAIK is let the user choose between:
1. Fast but (informatively) crashes in the event of an error.
2. Slow but you can recover from the error. 

Right now we can do that with only the check_errors flag. If you guys are ok then I will remove sync_exceptions.

The only option I see if we really do not want the code to crash is to let the results be silently wrong when check_errors is false, passing onto the user the responsibility to check for the num_pairs return value.

In my opinion this function should not let the code progress further if num_pairs> maximum_neighbors. The user is probably not going to bother checking and the danger of being silently wrong is not tolerable.
If the user wants to be able to do something about the exception (a.i. increase the maximum neighs) they can simply pass check_errors=True. Since recovering must be a conscious choice I believe this is a good behavior. 

For me, the ideal use case for this function would be something as follows:

```python

import torch as pt
from NNPOps.neighbors import getNeighborPairs

positions = pt.tensor(...)
max_num_neighbors = 1
# Find the maximum number of neighbors
while True:
    try:
        getNeighborPairs(positions, cutoff=3.0, max_num_neighbors=max_num_neighbors, check_errors=True)
    except RuntimeError:
        max_num_neighbors += 32
        continue
    break
# Fast and CUDA-graph compatible calls that should not ever raise, but will crash if an error occurs 
neigh, deltas, distances, num_pairs = getNeighborPairs(positions, cutoff=3.0, max_num_neighbors=max_num_neighbors)

```
But we can also make it be something like this:

```python

import torch as pt
from NNPOps.neighbors import getNeighborPairs

positions = pt.tensor(...)
max_num_neighbors = 1
# Find the maximum number of neighbors. This call will not ever raise, but be silently wrong.
neigh, deltas, distances, num_pairs = getNeighborPairs(positions, cutoff=3.0, max_num_neighbors=max_num_neighbors)
if num_pairs> max_num_neighbors:
    max_num_neighbors = num_pairs + 32
# This will also never raise, also silently wrong. This call and the above will be fast and CUDA-graph compatible
neigh, deltas, distances, num_pairs = getNeighborPairs(positions, cutoff=3.0, max_num_neighbors=max_num_neighbors)
# This will raise if necessary, but not be CUDA-graph compatible.
neigh, deltas, distances, num_pairs = getNeighborPairs(positions, cutoff=3.0, max_num_neighbors=max_num_neighbors, check_errors=True)
```
Let me know what you think!

peastman:
I don't think a crash is ever a good way of reporting an error.  I would vote for combining the flags so you have a single option.

- `check_errors=True`: Slower but you get an exception if it exceeds the maximum number of neighbors.
- `check_errors=False`: Faster but it's up to you to check the returned value to see if some neighbors were missed.

peastman:
> Then check_errors=True would be the default.

`check_errors=False` would be the default.

> In your proposal only check_errors=False would be compatible with CUDA graphs

Correct.

RaulPPelaez:
@peastman @raimis this is ready for review again.

src/pytorch/neighbors/getNeighborPairs.py
```diff
@@ -88,39 +104,39 @@ def getNeighborPairs(positions: Tensor, cutoff: float, max_num_neighbors: int =
...
     (tensor([[ 1, -1,  2],
              [ 0, -1,  1]], dtype=torch.int32),
      tensor([[1., 0., 0.],
              [nan, nan, nan],
              [1., 0., 0.]]),
-     tensor([1., nan, 1.]))
+     tensor([1., nan, 1.]), tensor([3], dtype=torch.int32))
 
     >>> getNeighborPairs(positions, cutoff=3.0, max_num_neighbors=2) # doctest: +NORMALIZE_WHITESPACE
     (tensor([[ 1,  2,  2, -1, -1, -1],
```
peastman, line 118: I don't understand the output.  If `max_num_neighbors` is 2, shouldn't the first return value have shape (2,2) instead of (2,6)?

There seems to be inconsistency in how we interpret `max_num_neighbors`.  In some places it's treated as the total number of neighbors in the list, but in others it's treated as the maximum number of neighbors for a single atom.

src/pytorch/neighbors/TestNeighbors.py
```diff
@@ -133,29 +133,69 @@ def test_neighbor_grads(device, dtype, num_atoms, grad):
...
         pytest.skip('No GPU')
-
     # 4 points result into 6 pairs, but there is a storage just for 4.
+    positions = pt.zeros((4, 3,), device=device, dtype=dtype)
     with pytest.raises(RuntimeError):
-        positions = pt.zeros((4, 3,), device=device, dtype=dtype)
-        getNeighborPairs(positions, cutoff=1, max_num_neighbors=1)
+        # checkErrors = True will raise due to exceeding neighbours
+        getNeighborPairs(positions, cutoff=1, max_num_neighbors=1, check_errors=True)
         pt.cuda.synchronize()
```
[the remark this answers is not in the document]
RaulPPelaez: Fixed

src/pytorch/neighbors/getNeighborPairs.py
```diff
@@ -63,18 +71,26 @@ def getNeighborPairs(positions: Tensor, cutoff: float, max_num_neighbors: int =
         If an atom pair is separated by a larger distance than the cutoff,
         the distance is set to `NaN`.
 
+    number_found_pairs: `torch.Tensor`
+        Contains the total  number of pairs found,  which might exceed
+        the  requested  max_num_neighbors,  leaving the  rest  of  the
+        output in an undefined state.
```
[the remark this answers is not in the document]
RaulPPelaez: What about this?: 
        Contains the total number of pairs found in an unspecified order,
        which might exceed the  requested  max_num_neighbors. In this case, the first number_found_pairs
        pairs in the output are valid pairs, but the remaining pairs are omitted.

src/pytorch/neighbors/getNeighborPairsCUDA.cu
```diff
@@ -64,14 +66,15 @@ template <typename scalar_t> __global__ void forward_kernel(
...
-    distances[i_pair] = sqrt_(distance2);
+    //We handle too many neighbors outside of the kernel
+    if(i_pair < neighbors.size(1)){
+      neighbors[0][i_pair] = row;
+      neighbors[1][i_pair] = column;
+      deltas[i_pair][0] = delta_x;
+      deltas[i_pair][1] = delta_y;
+      deltas[i_pair][2] = delta_z;
+      distances[i_pair] = sqrt_(distance2);
+    }
```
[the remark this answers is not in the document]
RaulPPelaez: Done, I usually run clang-format, but this file in particular has some formatting decisions that I have not been able to automate. So the formatter changes lines outside my changes... Would you consider giving #87 a spin?

src/pytorch/neighbors/getNeighborPairs.py
RaulPPelaez: You are right. max_num_neighbors is **per particle** as per the docs.  But only num_found_pairs<max_num_neighbors*num_atoms is enforced. 
For instance, if 4 particles are at the same point, there are 6 pairs, each particle having 3 neighbors.
However, the function will happily take max_num_neighbors=2, since that results in 8 maximum pairs.

The current build strategy does not provide an easy way to enforce max neighbors per particle.
OTOH changing the interface to receive maximum total pairs instead of per particle would make sense now, but could make it inconvenient if a different strategy is implemented in the future.
Still, the current output format (a compacted list of pairs) would probably require some kind of [partition](https://thrust.github.io/doc/group__partitioning_ga73ee335b49c6ae3dcb019423e268a231.html) if a more sophisticated algorithm would be in place.

Thus, I believe changing the meaning of the input parameter to "maximum number of total pairs" makes sense now.

I added a test that checks that the max_num_neighbors is actually enforced. Naturally, this test fails in cpu and gpu as of now.

So we have two options:
  1. Change the max_num_neighbors agument from "per particle" to max_num_pairs.
  2. Simply change the documentation to reflect that, while an atom might have more than max_num_neighbors, this is not an error unless num_pairs>max_num_neighbors*num_atoms

What do you think?
cc @raimis

src/pytorch/neighbors/TestNeighbors.py
```diff
@@ -155,6 +155,19 @@ def test_too_many_neighbors(device, dtype):
...
+    if not pt.cuda.is_available() and device == 'cuda':
+        pytest.skip('No GPU')
+    # 4 points result into 6 pairs.
+    positions = pt.zeros((4, 3,), device=device, dtype=dtype)
+    with pytest.raises(RuntimeError):
+        # checkErrors = True should raise due to exceeding neighbours
+        # As of now this will not raise, since 6<num_atoms*max_num_neighbors=8
+        getNeighborPairs(positions, cutoff=1, max_num_neighbors=2, check_errors=True)
+    #This should not fail, there are 4 neighbors per particle
+    getNeighborPairs(positions, cutoff=1, max_num_neighbors=4, check_errors=True)
```
RaulPPelaez: This is the failing test.
What should we do about this inconsistency between the interface and the implementation?

src/pytorch/neighbors/getNeighborPairs.py
peastman: Given how the code is written, I think a limit on the total pairs is what we need.  The reason for having this argument at all is that we need to know what size to make the output tensors.  And nothing in the output tensors depends on how many neighbors a particular atom has.  They just depend on the total number of pairs in the neighbor list.

src/pytorch/neighbors/getNeighborPairs.py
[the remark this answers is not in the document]
peastman: How about this?  "Contains the total number of pairs found.  Be aware that if `check_errors` is False, this might be larger than `max_num_neighbors`.  In that case, the output tensors contain only a subset of the pairs that were found, and the others are omitted.  Which pairs get omitted may vary between invocations."

src/pytorch/neighbors/getNeighborPairs.py
RaulPPelaez: I changed the meaning and name from max_num_neighbors (maximum number of neighbors per particle) to max_num_pairs (maximum number of total pairs).
I modified the tests to reflect this.

RaulPPelaez:
This is ready for review again.

raimis:
Ping @peastman

peastman:
This addresses all the issues I raised.  Looks good to me now!

src/pytorch/neighbors/getNeighborPairs.py
```diff
@@ -63,17 +74,27 @@ def getNeighborPairs(positions: Tensor, cutoff: float, max_num_neighbors: int =
         If an atom pair is separated by a larger distance than the cutoff,
         the distance is set to `NaN`.
 
+    number_found_pairs: `torch.Tensor`
+        Contains the  total number  of pairs found.  Be aware  that if
+        check_errors   is   False,   this   might   be   larger   than
+        max_num_pairs. In  that case,  the output  tensors contain
```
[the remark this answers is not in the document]
RaulPPelaez: Yes, I usually make paragraphs justified within 70 cols for large comments. I use C-u 1 M-q in emacs, runs fill-paragraph. For instance, this:
```python
'''
        Atom pair displacement vectors. The shape of the tensor is `(num_pairs, 3)`.
        The direction of the vectors are from `neighbors[1]` to `neighbors[0]`.
        If an atom pair is separated by a larger distance than the cutoff,
        the displacement vector is set to `[NaN, NaN, NaN]`.
'''
```
Becomes this:
```python
'''
        Atom pair  displacement vectors.  The shape  of the  tensor is
        `(num_pairs,  3)`.   The direction  of  the  vectors are  from
        `neighbors[1]`  to   `neighbors[0]`.   If  an  atom   pair  is
        separated  by   a  larger   distance  than  the   cutoff,  the
        displacement vector is set to `[NaN, NaN, NaN]`.
'''
```
I can stop doing that if you'd prefer.

raimis:
@RaulPPelaez I suppose, this is ready to merge?

RaulPPelaez:
This is ready for merge.

PR #81, opened 2023-01-17
Now CMakeFiles.txt only uses find_package(Python), not PythonLibs

RaulPPelaez:
Calling find_package with Python and then PythonLibs was finding the python exec from conda but the libs and includes in my system, which was causing all kinds of weird errors.
find_package(PythonLibs) [was deprecated in CMake 3.12](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html), find_package(Python) supersedes it. The current CMakeLists requires 3.20.
I modified CMakeLists.txt to only use find_package(Python) for the exec and the development libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,8 +9,7 @@ if(ENABLE_CUDA)
 endif(ENABLE_CUDA)
 
 # Find dependencies
-find_package(Python REQUIRED)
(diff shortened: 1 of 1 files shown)


raimis (approved):
Looks good!

raimis:
Rename to make a release log clear.

PR #83, opened 2023-02-27
Fix PBCs

### LINKED ISSUE #82: Surprising differences between nnpops and torchani
Hi,

I have been getting surprising results running waterbox simulations with the `torchani` vs the `nnpops` implementation of `ani2x`. I used the `openmmtools` waterbox testsystem with an edge length of 20 A, and a 1 fs timestep, and simulated for 1 ns using a Langevin integrator with 1/ps collision rate. The system was set up with `potential.createSystem()`.

When I run simulations in NpT with the `torchani` implementation at 300 K everything looks relatively normal (density is a bit too high, and the rdf has some surprising signal though):
image
When I perform the same simulation with the `nnpops` implementation I see this:
image
and the simulation box has shrunk (the initial box size is the yellow outlined square). Also, note the difference in the y-axis for the potential energy.
image

In NVT I observe vacuum bubbles with `nnpos`
https://user-images.githubusercontent.com/31651017/218335894-0254ed80-e51f-4189-9bfc-ae94637cfd85.mp4

compared to the same simulation with `torchani`
https://user-images.githubusercontent.com/31651017/218335817-3e911757-d19d-4f71-b922-8b9de913237e.mp4

I attached a minimal example to reproduce the simulations.

[min_example.py.zip](https://github.com/openmm/NNPOps/files/10717263/min_example.py.zip)

sef43:
This aims to fix #82 
It sets the `periodic` flag  in SymmetryFunctions.cpp based on if the cell vectors are defined as suggested by @peastman 

I have added a new test case that is a water box with PBCs, this is run in the TestOptimizedTorchANI.py and TestSymmetryFunctions.py tests where the energy and forces of the NNPOps implementation are compared to the TorchANI implementation.


peastman:
Looks good to me.  All tests have passed.  I'll leave it to @raimis to merge if it looks ok to him.

raimis:
@sef43 thanks for the fix!

raimis:
Rename to make a release log clear.

PR #88, opened 2023-03-03
Update CI versions

raimis:
- [ ] Update to PyTorch 1.13

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,20 +41,20 @@ jobs:
             pytorch: "1.11.*"
 
           # Latest supported versions (with CUDA)
-          - name: Linux (CUDA 11.7, Python 3.10, PyTorch 1.12)
+          - name: Linux (CUDA 11.7, Python 3.10, PyTorch 1.13)
             enable_cuda: true
             cuda: "11.7.0"
             gcc: "10.3.*"
             nvcc: "11.7"
             python: "3.10.*"
-            pytorch: "1.12.*"
+            pytorch: "1.13.*"
 
           # Latest supported versions (without CUDA)
-          - name: Linux (no CUDA, Python 3.10, PyTorch 1.12)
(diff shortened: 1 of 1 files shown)


peastman:
Why drop CUDA 10.2?

raimis:
CUDA 10.2 was dropped by conda-forge (https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/3491)

peastman:
That just says the PyTorch maintainers are considering dropping it for future PyTorch builds on conda-forge.  That doesn't mean we need to stop testing it.  Especially since we're still testing older PyTorch versions that provide CUDA 10.2 builds.

raimis:
It was dropped two months ago (https://github.com/conda-forge/nnpops-feedstock/commit/c983beb33f11cf2bb4d64070aa4ee67ed9553b86)

raimis:
@RaulPPelaez @sef43 do you see any reason to keep testing on CUDA 10.2?

sef43:
> @RaulPPelaez @sef43 do you see any reason to keep testing on CUDA 10.2?

I think It seems reasonable to drop

RaulPPelaez:
I believe there is value in leaving old versions there. If and when CUDA 10.2 tests do not pass then we can decide if its worth to do something about it or simply drop it to reduce maintainer burden as pytorch just did.

raimis:
OK! Let's keep CUDA 10.2 until it works.

RaulPPelaez:
> @peastman @RaulPPelaez @sef43 could you review?

torch 1.13 resulted in some tests failing, I believe that is not solved yet.

raimis:
Keep in mind, CI doesn't run the tests on GPU.

RaulPPelaez:
Oh right. Looks fine to me then.

peastman (approved):
Looks good to me.

PR #89, opened 2023-03-07
Move all torch.ops.load calls to the __init__.py scripts

### LINKED ISSUE #68: getNeighborPairs error
I am trying to use the new neighbourlist with pytorch 1.12.1. I have managed to install `NNPOps` via conda. 

If I check the PyTorch version it is 1.12, so that is fine. 

If I try to import NNPOps that is also fine:

```sh
>>> from NNPOps.neighbors import getNeighborPairs
```

And it is event telling me it needs two positional arguments:

```sh
>>> getNeighborPairs()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: getNeighborPairs() missing 2 required positional arguments: 'positions' and 'cutoff'
```

But when I actually specify those positional arguments I am getting the following error:

```sh
>>> getNeighborPairs(positions=positions, cutoff=cutoff)
Traceback (most recent call last):
  File "/home/dpk25/.conda/envs/MACE/lib/python3.10/site-packages/torch/_ops.py", line 198, in __getattr__
    op, overload_names = torch._C._jit_get_operation(qualified_op_name)
RuntimeError: No such operator neighbors::getNeighborPairs

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dpk25/.conda/envs/MACE/lib/python3.10/site-packages/NNPOps/neighbors/getNeighborPairs.py", line 82, in getNeighborPairs
    return ops.neighbors.getNeighborPairs(positions, cutoff, max_num_neighbors)
  File "/home/dpk25/.conda/envs/MACE/lib/python3.10/site-packages/torch/_ops.py", line 202, in __getattr__
    raise AttributeError(f"'_OpNamespace' object has no attribute '{op_name}'") from e
AttributeError: '_OpNamespace' object has no attribute 'getNeighborPairs'
```

Does anyone perhaps know what might be going wrong? 

A few relevant info about the environment:

```sh
python                    3.10.4               h12debd9_0  
torch                     1.12.1+cu113             pypi_0    pypi
```

Interestingly, `NNPOps` does not appear when I do `conda list`.

RaulPPelaez:
This PR tries to fix #68 by moving all calls to torch.ops.load to the __init__.py scripts.

### DIFF
--- a/src/pytorch/__init__.py
+++ b/src/pytorch/__init__.py
@@ -12,7 +12,7 @@ for path in site.getsitepackages():
         break
 else:
     # if we didn't find it, look for NNPOps/libNNPOpsPyTorch.so in the same directory as this file
-    torch.ops.load_library(os.path.join(os.path.dirname(os.path.dirname(__file__)), 'libNNPOpsPyTorch.so'))
+    torch.ops.load_library(os.path.join(os.path.dirname(__file__), 'libNNPOpsPyTorch.so'))
 
 
 from NNPOps.OptimizedTorchANI import OptimizedTorchANI
--- a/src/pytorch/__init__.py
+++ b/src/pytorch/__init__.py
@@ -2,9 +2,9 @@
 High-performance PyTorch operations for neural network potentials
 '''
 import os.path
-import site
 import torch
-torch.ops.load_library(os.path.join(site.getsitepackages()[-1],"NNPOps", "libNNPOpsPyTorch.so"))
-torch.classes.load_library(os.path.join(site.getsitepackages()[-1],"NNPOps", "libNNPOpsPyTorch.so"))
+
+torch.ops.load_library(os.path.join(os.path.dirname(__file__), 'libNNPOpsPyTorch.so'))
+torch.classes.load_library(os.path.join(os.path.dirname(__file__), 'libNNPOpsPyTorch.so'))
 
 from NNPOps.OptimizedTorchANI import OptimizedTorchANI
--- a/src/pytorch/__init__.py
+++ b/src/pytorch/__init__.py
@@ -3,8 +3,16 @@ High-performance PyTorch operations for neural network potentials
 '''
 import os.path
 import torch
+import site
 
-torch.ops.load_library(os.path.join(os.path.dirname(__file__), 'libNNPOpsPyTorch.so'))
-torch.classes.load_library(os.path.join(os.path.dirname(__file__), 'libNNPOpsPyTorch.so'))
+# look for NNPOps/libNNPOpsPyTorch.so in all the paths returned by site.getsitepackages()
+for path in site.getsitepackages():
+    if os.path.exists(os.path.join(path, 'NNPOps/libNNPOpsPyTorch.so')):
+        torch.ops.load_library(os.path.join(path, 'NNPOps/libNNPOpsPyTorch.so'))
+        break
+else:
+    # if we didn't find it, look for NNPOps/libNNPOpsPyTorch.so in the same directory as this file
+    torch.ops.load_library(os.path.join(os.path.dirname(os.path.dirname(__file__)), 'libNNPOpsPyTorch.so'))
+
 
 from NNPOps.OptimizedTorchANI import OptimizedTorchANI
--- a/src/pytorch/__init__.py
+++ b/src/pytorch/__init__.py
@@ -3,16 +3,8 @@ High-performance PyTorch operations for neural network potentials
 '''
 import os.path
 import torch
-import site
 
-# look for NNPOps/libNNPOpsPyTorch.so in all the paths returned by site.getsitepackages()
-for path in site.getsitepackages():
-    if os.path.exists(os.path.join(path, 'NNPOps/libNNPOpsPyTorch.so')):
-        torch.ops.load_library(os.path.join(path, 'NNPOps/libNNPOpsPyTorch.so'))
-        break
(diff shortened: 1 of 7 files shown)


raimis:
@RaulPPelaez is this ready for review?

RaulPPelaez:
Yes, thanks.

src/pytorch/neighbors/__init__.py
```diff
@@ -1,5 +1,10 @@
 '''
 Neighbor operations
 '''
+import site
+import os
+import torch
 
-from NNPOps.neighbors.getNeighborPairs import getNeighborPairs
\ No newline at end of file
+torch.ops.load_library(os.path.join(site.getsitepackages()[-1],"NNPOps", "libNNPOpsPyTorch.so"))
```
raimis, line 8: Does this really need? I think `src/pytorch/__init__.py` is executed when `NNPOps.neighbors` is imported.

src/pytorch/__init__.py
```diff
@@ -1,5 +1,10 @@
 '''
 High-performance PyTorch operations for neural network potentials
 '''
+import os.path
+import site
+import torch
+torch.ops.load_library(os.path.join(site.getsitepackages()[-1],"NNPOps", "libNNPOpsPyTorch.so"))
```
raimis, line 7: `site.getsitepackages` returns a list of paths. I think, we check all of them for `NNPOps/libNNPOpsPyTorch.so` and load the first found.

src/pytorch/neighbors/__init__.py
RaulPPelaez: This example does not run for me unless I load the library there:
```python
import torch as pt                                                                                                                                                           
import numpy as np                                                                                                                                                           
import pytest                                                                                                                                                                
from NNPOps.neighbors import getNeighborPairs                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
device = 'cuda:0'
dtype = pt.float32
num_atoms = 100
# Generate random positions
positions = 10 * pt.randn((num_atoms, 3), device=device, dtype=dtype)
cutoff = float(5.0)
num_neighbors = num_atoms*num_atoms
neighbors, deltas, distances = getNeighborPairs(positions, cutoff=cutoff, max_num_neighbors=num_neighbors)
```
Seems like if one includes NNPOps.neighbors directly as above the pytorch/__init__.py is not loaded.

src/pytorch/neighbors/__init__.py
```diff
@@ -1,5 +1,9 @@
 '''
 Neighbor operations
 '''
+import os.path
+import torch
 
-from NNPOps.neighbors.getNeighborPairs import getNeighborPairs
\ No newline at end of file
+torch.ops.load_library(os.path.join(os.path.dirname(os.path.dirname(__file__)), 'libNNPOpsPyTorch.so'))
```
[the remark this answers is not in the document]
RaulPPelaez: This was intended, as libNNPOps is at the parent directory with respect to neighbors.
In any case, this line sometimes tries to look in the src folder instead of the installation one. I thus went back to your idea of looking for the library in all the folders returned by getsitepackages().

Now tests pass on my machine and the original issue is fixed. Lets wait for the CI...

src/pytorch/neighbors/__init__.py
RaulPPelaez: Then we should be able to remove the torch.load in neighbors. Lets see.

RaulPPelaez:
This is ready for review.

src/pytorch/__init__.py
```diff
@@ -1,5 +1,18 @@
...
 High-performance PyTorch operations for neural network potentials
 '''
+import os.path
+import torch
+import site
 
-from NNPOps.OptimizedTorchANI import OptimizedTorchANI
\ No newline at end of file
+# look for NNPOps/libNNPOpsPyTorch.so in all the paths returned by site.getsitepackages()
+for path in site.getsitepackages():
```
raimis, line 9: We talked about using only the relative paths to find the library. `site.getsitepackages` is unnecessary, because we create the package in a specific way.

RaulPPelaez: At installation __init__.py is located at the same place as libNNPOpsPyTorch.so, so we can simply use:

```python
torch.ops.load_library(os.path.join(os.path.dirname(__file__), 'libNNPOpsPyTorch.so'))
```

However, this breaks testing, since tests are located along the source files and make test runs lines like:
https://github.com/openmm/NNPOps/blob/293d90826371a20e63fa3bb21880cfd10132198a/CMakeLists.txt#L60

I believe this loads ```src/pytorch/__init__.py```, but since ```libNNPOpsPyTorch.so``` lives at sitepackages (and the build folder) the load fails. 

I only see two solutions to this:
1. Take the test scripts to another place (a test directory at the root?)
2. Modify the CMakeLists.txt to copy the test scripts to the build directory and run them from there.

2 seems like the best compromise to me. What do you think?

peastman: > 2 seems like the best compromise to me. What do you think?

I agree.  That's what OpenMM does.

raimis (approved):
Looks good!

raimis:
@peastman any comments? Otherwise, we can merge.

peastman:
Looks good to me.

PR #94, opened 2023-03-17
Support for pytorch 2.0

RaulPPelaez:
This PR is to start working on making NNPOps compatible with pytorch 2.0 and torch.compile

- [x] Ensure the workflows and tests work for pytorch 2.0 (wich requires CUDA 11.8)
- [ ] Make operations compatible with torch.compile()
   Hopefully this will not require a lot of work, but reducing the number of graph breaks torch.compile will introduce will probably be more challenging. I reckon avoiding graph breaks is really similar to making stuff compatible with CUDA graphs. 
- [ ] Write tests for compiled versions
       In principle compiled functions/models should be completely equivalent to the uncompiled versions, but I have seen this not being the case (granted, torch2 was still a beta)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
           # Latest supported versions (with CUDA)
           - name: Linux (CUDA 11.8, Python 3.10, PyTorch 2.0)
             enable_cuda: true
-            cuda: "11.8"
+            cuda: "11.8.0"
             gcc: "10.3.*"
             nvcc: "11.8"
             python: "3.10.*"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,9 +41,9 @@ jobs:
             pytorch: "1.12.*"
 
           # Latest supported versions (with CUDA)
-          - name: Linux (CUDA 11.7, Python 3.10, PyTorch 2.0)
+          - name: Linux (CUDA 11.8.0, Python 3.10, PyTorch 2.0)
             enable_cuda: true
-            cuda: "11.7.*"
+            cuda: "11.8.0"
             gcc: "10.3.*"
             nvcc: "11.7.0"
             python: "3.10.*"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,9 +41,9 @@ jobs:
             pytorch: "1.12.*"
 
           # Latest supported versions (with CUDA)
-          - name: Linux (CUDA 11.7, Python 3.10, PyTorch 2.0)
+          - name: Linux (CUDA 11.8, Python 3.10, PyTorch 2.0)
             enable_cuda: true
-            cuda: "11.7.0"
+            cuda: "11.8.*"
             gcc: "10.3.*"
             nvcc: "11.7.0"
             python: "3.10.*"
@@ -54,7 +54,7 @@ jobs:
             enable_cuda: false
             gcc: "10.3.*"
             python: "3.10.*"
-            pytorch: "1.13.*"
+            pytorch: "2.*"
 
     steps:
     - name: Check out
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Install CUDA Toolkit
-      uses: Jimver/cuda-toolkit@v0.2.8
+      uses: Jimver/cuda-toolkit@v0.2.10
       with:
         cuda: ${{ matrix.cuda }}
         linux-local-args: '["--toolkit", "--override"]' 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,13 +41,13 @@ jobs:
             pytorch: "1.12.*"
 
           # Latest supported versions (with CUDA)
-          - name: Linux (CUDA 11.7, Python 3.10, PyTorch 1.13)
+          - name: Linux (CUDA 11.8, Python 3.10, PyTorch 2.0)
             enable_cuda: true
-            cuda: "11.7.0"
+            cuda: "11.8.*"
             gcc: "10.3.*"
-            nvcc: "11.7"
+            nvcc: "11.8.*"
             python: "3.10.*"
-            pytorch: "1.13.*"
+            pytorch: "2.*"
 
           # Latest supported versions (without CUDA)
           - name: Linux (no CUDA, Python 3.10, PyTorch 1.13)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
       run: |
         sed -i -e "/cudatoolkit/c\  - cudatoolkit ${{ matrix.cuda }}" \
                -e "/gxx_linux-64/c\  - gxx_linux-64 ${{ matrix.gcc }}" \
-               -e "/torchani/c\  - torchani ${{ matrix.gcc }}" \
+               -e "/torchani/c\  - torchani ${{ matrix.torchani }}" \
                -e "/nvcc_linux-64/c\  - nvcc_linux-64 ${{ matrix.nvcc }}" \
                -e "/python/c\  - python ${{ matrix.python }}" \
                -e "/pytorch-gpu/c\  - pytorch-gpu ${{ matrix.pytorch }}" \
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -1,14 +1,16 @@
 channels:
   - conda-forge
+  - pytorch
```
sef43, line 3: I think we will need to wait for the conda-forge pytorch to be updated to 2.0?

RaulPPelaez: Pytorch's web suggest this for installing pytorch 2:
```bash
conda install pytorch torchvision torchaudio pytorch-cuda=11.8 -c pytorch -c nvidia
```
Do you think adding the additional channels will be a problem?

sef43: As I understand it to release on conda-forge all dependencies need to be available from the conda-forge channel: https://conda-forge.org/docs/maintainer/adding_pkgs.html#avoid-external-dependencies

RaulPPelaez: I see, then I guess next release will have to wait for that... We can work on making everything compatible for that moment in the meanwhile.

RaulPPelaez:
Torchani cannot be installed with pytorch2, which forces to skip some tests.

RaulPPelaez:
All tests pass and the ci works for an installation with pytorch2. I believe this should be merged now and work on compile() compatibility be done in another PR.
A new release could be done now so that users can install NNPOps along pytorch2.

RaulPPelaez:
In case you have some experience with torch2 compile:
This test miserably fails in CUDA mode:
```python

@pytest.mark.parametrize('device', ['cpu', 'cuda'])
@pytest.mark.parametrize('dtype', [pt.float32, pt.float64])
def test_torch_compile_compatible(device, dtype):

    class ForceModule(pt.nn.Module):

        def forward(self, positions):

            neighbors, deltas, distances = getNeighborPairs(positions, cutoff=1.0)
            mask = pt.isnan(distances)
            distances = distances[~mask]
            return pt.sum(distances**2)

    original_model = ForceModule()
    num_atoms=10
    positions = (20 * pt.randn((num_atoms, 3), device=device, dtype=dtype)) - 10
    original_model(positions)
    model = pt.compile(original_model)
    model(positions)
```
It yields a really verbose error about something called FakeTensor that makes the most obscure gcc recursive template error look clear and informative:

<details>

```python

TestNeighbors.py::test_torch_compile_compatible[dtype1-cuda] FAILED                                                                                                [600/1860]
                                                                                                                                                                             
================================================================================= FAILURES ==================================================================================
________________________________________________________________ test_torch_compile_compatible[dtype0-cuda] _________________________________________________________________
                                                                                                                                                                             
output_graph = <torch._dynamo.output_graph.OutputGraph object at 0x7fc3e7d81fc0>, node = get_neighbor_pairs                                                                  
args = (FakeTensor(FakeTensor(..., device='meta', size=(10, 3)), cuda:0), 1.0, -1, FakeTensor(FakeTensor(..., device='meta', size=(0, 0)), cuda:0)), kwargs = {}             
nnmodule = None                                                                                                                                                              
                                                                                                                                                                             
    def run_node(output_graph, node, args, kwargs, nnmodule):                                                                                                                
        """                                                                                                                                                                  
        Runs a given node, with the given args and kwargs.
     
        Behavior is dicatated by a node's op.
     
        run_node is useful for extracting real values out of nodes.
        See get_real_value for more info on common usage.
     
        Note: The output_graph arg is only used for 'get_attr' ops
        Note: The nnmodule arg is only used for 'call_module' ops
     
        Nodes that are not call_function, call_method, call_module, or get_attr will
        raise an AssertionError.
        """
        op = node.op
        try:
            if op == "call_function":
>               return node.target(*args, **kwargs)

../../../mambaforge/envs/nnpops-torch2-nvidia/lib/python3.10/site-packages/torch/_dynamo/utils.py:1194: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <OpOverloadPacket(op='neighbors.getNeighborPairs')>
args = (FakeTensor(FakeTensor(..., device='meta', size=(10, 3)), cuda:0), 1.0, -1, FakeTensor(FakeTensor(..., device='meta', size=(0, 0)), cuda:0)), kwargs = {}

    def __call__(self, *args, **kwargs):
        # overloading __call__ to ensure torch.ops.foo.bar()
        # is still callable from JIT
        # We save the function ptr as the `op` attribute on
        # OpOverloadPacket to access it here.
>       return self._op(*args, **kwargs or {})
E       RuntimeError: The tensor has a non-zero number of elements, but its data is not allocated yet. Caffe2 uses a lazy allocation, so you will need to call mutable_data()
 or raw_mutable_data() to actually allocate memory.

../../../mambaforge/envs/nnpops-torch2-nvidia/lib/python3.10/site-packages/torch/_ops.py:502: RuntimeError

```
and following for a gazillion lines.
</details>

I have not been able to solve this, from what I have gathered this should not happen and it is a bug in torch (there are a lot of issues describing stuff like this: https://github.com/pytorch/pytorch/issues/96742 https://github.com/pytorch/pytorch/issues/95791

raimis:
Yes, we can skip the compile feature for now.

.github/workflows/ci.yml
```diff
@@ -38,33 +39,36 @@ jobs:
...
 
           # Latest supported versions (with CUDA)
-          - name: Linux (CUDA 11.7, Python 3.10, PyTorch 1.13)
+          - name: Linux (CUDA 11.8.0, Python 3.10, PyTorch 2.0)
             enable_cuda: true
-            cuda: "11.7.0"
+            cuda: "11.8.0"
             gcc: "10.3.*"
-            nvcc: "11.7"
+            nvcc: "11.7.0"
```
raimis, line 50: This should be `11.8` (https://anaconda.org/conda-forge/nvcc_linux-64/files)

.github/workflows/ci.yml
```diff
@@ -38,33 +39,36 @@ jobs:
...
             python: "3.10.*"
-            pytorch: "1.13.*"
+            torchani: "2.2.*"
+            pytorch: "2.*"
```
raimis, line 53: We should pin to `2.0.*`

.github/workflows/ci.yml
```diff
@@ -38,33 +39,36 @@ jobs:
...
 
           # Latest supported versions (without CUDA)
           - name: Linux (no CUDA, Python 3.10, PyTorch 1.13)
             enable_cuda: false
             gcc: "10.3.*"
             python: "3.10.*"
-            pytorch: "1.13.*"
+            pytorch: "2.*"
```
raimis, line 60: We should pin to `2.0.*`

.github/workflows/ci.yml
```diff
@@ -38,33 +39,36 @@ jobs:
...
             gcc: "10.3.*"
-            nvcc: "11.7"
+            nvcc: "11.7.0"
             python: "3.10.*"
-            pytorch: "1.13.*"
+            torchani: "2.2.*"
+            pytorch: "2.*"
 
           # Latest supported versions (without CUDA)
           - name: Linux (no CUDA, Python 3.10, PyTorch 1.13)
```
raimis, line 56: PyTorch 2.0

environment.yml
```diff
@@ -2,13 +2,13 @@ channels:
...
-  - nvcc_linux-64 11.2
-  - torchani 2.2.2
+  - nvcc_linux-64 11.8.*
+  - torchani =2.2.3
   - pytest
   - python 3.10.*
-  - pytorch-gpu 1.12.*
-  - sysroot_linux-64 2.17
\ No newline at end of file
+  - pytorch-gpu 2.*
```
raimis, line 13: We should pin to `2.0.*`

environment.yml
```diff
@@ -2,13 +2,13 @@ channels:
...
 dependencies:
   - cmake >=3.20
-  - cudatoolkit 11.2.2
+  - cudatoolkit 11.8.*
   - gxx_linux-64 10.3.*
   - make
   - mdtraj
-  - nvcc_linux-64 11.2
-  - torchani 2.2.2
+  - nvcc_linux-64 11.8.*
```
raimis, line 9: `11.8`

environment.yml
```diff
@@ -2,13 +2,13 @@ channels:
...
+  - torchani =2.2.3
```
raimis, line 10: `2.2.*` to be consistent with CI

RaulPPelaez:
Ok I think this is done now.

.github/workflows/ci.yml
```diff
@@ -38,33 +39,36 @@ jobs:
...
             python: "3.9.*"
+            torchani: "2.2.*"
             pytorch: "1.12.*"
 
           # Latest supported versions (with CUDA)
-          - name: Linux (CUDA 11.7, Python 3.10, PyTorch 1.13)
+          - name: Linux (CUDA 11.8, Python 3.10, PyTorch 2.0)
             enable_cuda: true
-            cuda: "11.7.0"
+            cuda: "11.8.0"
```
raimis, line 48: I think it should be `11.8.*`. Bugfixes shouldn't break API.

raimis: No, it is fine. The CUDA installer needs a complete version.

raimis (approved):

raimis:
@RaulPPelaez can I merge?

PR #97, opened 2023-03-23
Fix CMake not updating tests correctly after first run

RaulPPelaez:
The current rule that copies tests to the build folder does not take into account the test scripts changing.
I modified the cmake command so that it copies the scripts every time make is called.


RaulPPelaez:
This is ready for merging

raimis (approved):
Looks good!

Issue #98, opened 2023-03-31
Many tests are being skipped by the CI


RaulPPelaez:
The CI runs this to test:
```shell
ctest --verbose --exclude-regex TestCuda
```
Running this on my machine results in 11 tests. But the CI only runs 3.
I cannot find out why.

mikemhenry:
It looks like we run 5 on the GPU CI:
https://github.com/openmm/NNPOps/actions/runs/5063145679/jobs/9089458222#step:9:75
which just calls `ctest --verbose`

RaulPPelaez:
I think I know whats going on:
https://github.com/openmm/NNPOps/blob/054d487d9fec8f98a111bc30ec6d3ec1ce423356/CMakeLists.txt#L60-L77

This assumes that the custom target runs before the loop in line 74. Since this is not the case, the first time one runs cmake there is no file under CMAKE_BINARY_DIR/test and thus the tests are skipped.
I just tried and confirmed this is the case, run cmake once and you get 3 tests, run twice and you get 11.

RaulPPelaez:
I just want to point out that I gave the above snippet to gpt4 without any context and the thing answered this:

> There seems to be an issue here, though. The add_test command is run at configure time, but the files that you want to glob won't exist until build time, after the custom command is run. Therefore, file(GLOB_RECURSE PYTHON_TEST_PATHS ${CMAKE_BINARY_DIR}/test/Test*.py) will return an empty list.
> 
> To fix this issue, you could move the test discovery and registration into a separate script that you call using add_test, or consider configuring the tests at build time. This is a known limitation of CMake's test discovery.

Dang...
I opened #104 to work on this.

PR #100, opened 2023-04-19
Add Self-Hosted AWS GPU Runner

mikemhenry:


environment.yml
```diff
@@ -2,13 +2,10 @@ channels:
   - conda-forge
 dependencies:
   - cmake >=3.20
-  - cudatoolkit 11.8.*
```
raimis, line 5: Don't remove dependencies. The file should work for the build from source too (https://github.com/openmm/NNPOps#build-from-source)

mikemhenry: I'm in the middle of playing with this, I got AWS GPU stuff working :sunglasses: My plan is once it is all working, move the env file for CI to either a different folder or name, so that the installation instructions are still correct, but it will make changing the build matrix easier using `extra-specs` instead of `sed` to modify the environment file on the fly

raimis: I would prefer to use a single environment file and modify it as needed. This ensures consistency.

.github/workflows/ci.yml
```diff
@@ -71,52 +71,45 @@ jobs:
         linux-local-args: '["--toolkit", "--override"]' 
       if: ${{ matrix.enable_cuda }}
 
-    - name: Install Miniconda
```
raimis, line 74: Could this PR be limited to adding the GPU runner rather than changing entire CI?

mikemhenry: Sure, I was just trying to save some CI time and make how I did the GPU runner match the rest of the CI, I will make as small of a change as I can

raimis: Thanks for trying to bring innovations. Showcase them with the GPU CI and we will consider updating the rest CI.

mikemhenry:
Ready for review @RaulPPelaez and @raimis 

I only tested the latest version and before we merge this in we will want to decide how often to run the scheduled test and we want to remove the part the triggers the run in this branch.

mikemhenry:
Also one more thing I want to do before merge is to use variables for he package versions so it is easier to change in the future

environment.yml
RaulPPelaez: I believe we should switch to mamba-org if it is faster.

.github/workflows/self-hosted-gpu-test.yml
```diff
@@ -0,0 +1,126 @@
...
+          
+    - name: Prepare dependencies (with CUDA)
+      run: |
+        sed -i -e "/cudatoolkit/c\  - cudatoolkit 11.7.*" \
+               -e "/gxx_linux-64/c\  - gxx_linux-64 10.3.*" \
+               -e "/torchani/c\  - torchani 2.2*" \
+               -e "/nvcc_linux-64/c\  - nvcc_linux-64 11.7.*" \
+               -e "/python/c\  - python 3.10.*" \
+               -e "/pytorch-gpu/c\  - pytorch-gpu 2.0.*" \
+               environment.yml
```
RaulPPelaez, line 75: This should somehow use the same values as the current ci.yml, even if it is only for the latest versions.
Do you guys think it is worth joining this yml and the current ci.yml?

.github/workflows/self-hosted-gpu-test.yml
```diff
@@ -0,0 +1,126 @@
...
+            [
+              {"Key": "Name", "Value": "ec2-github-runner"},
+              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
+            ]
+
+  do-the-job:
+    name: Do the job on the runner
+    needs: start-runner # required to start the main job when the runner is ready
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    timeout-minutes: 1200 # 20 hrs
```
RaulPPelaez, line 51: I would say if this takes more than 1 hour then the test should be considered failing because something is clearly going on.
Is there some reason why you put 20h?

.github/workflows/self-hosted-gpu-test.yml
```diff
@@ -0,0 +1,126 @@
+name: self-hosted-gpu-test
+on:
+  push:
+    branches:
+      - master
+      - feat/add_aws_testing
```
RaulPPelaez, line 6: This should not be here, right?

RaulPPelaez:
This is a really cool functionality @mikemhenry, thanks! Its just some minor comments from me.

.github/workflows/self-hosted-gpu-test.yml
```diff
@@ -0,0 +1,126 @@
...
+    - cron: "0 0 * * SUN"
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+  start-runner:
+    name: Start self-hosted EC2 runner
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
mikemhenry: Not a bad idea generally, but this `runs-on` just launches the EC2 runner, so we don't actually run anything on it.

.github/workflows/self-hosted-gpu-test.yml
mikemhenry: Is the environment yaml that is in the root of the repo an example of the "latest" verions of pins? I was going to refactor this a bit like I did here: https://github.com/openmm/openmm-torch/pull/107/files#diff-bd64c4de8527b7011b64ffef8a82f8ab32dae8c0ac01c59a71d4f3b28f05cfddR47-R54 so that it will be easy to modify these and update to the latest versions, much like how the github CI is setup.

.github/workflows/self-hosted-gpu-test.yml
mikemhenry: Haha, I copied this from another repo where we run some long tests on AWS. These much faster so I can set the timeout to be shorter, great catch!

.github/workflows/self-hosted-gpu-test.yml
mikemhenry: Yes, I will remove that before we merge it. For testing/making this PR, this makes it easy to test.

.github/workflows/self-hosted-gpu-test.yml
RaulPPelaez: I was thinking about this:
https://github.com/openmm/NNPOps/blob/b63fc70e9b671988f67aec1fbc62bbd01f830b25/.github/workflows/ci.yml#L22-L61
Which transforms the env yaml like this:
https://github.com/openmm/NNPOps/blob/b63fc70e9b671988f67aec1fbc62bbd01f830b25/.github/workflows/ci.yml#L83-L90
So basically what you mentioned.

mikemhenry:
Okay this is ready for review @raimis and @RaulPPelaez 
The only thing left is to decide how frequent you want to run it.
It costs @jchodera $0.526 per hour, and it looks like it runs for ~ 11 minutes so it is pretty cheap. I think running it on every commit on a PR though is wasteful. I recommend running it nightly, merge commits to master, and manually on a branch that can be done on-demand (useful to do before merging).

peastman:
If we want to keep the cost down, running it only on merge commits to master is probably sufficient.  Running it nightly on code that's already been merged and tested doesn't tell us much more.

RaulPPelaez:
I agree with @peastman, nightly is too much. I say we run this just before committing to master.
Other than that the ci looks great to me, if we want to play it safe get the timeout down to 25 mins.
Thanks @mikemhenry

mikemhenry:
Okay! This should be good to merge now!

RaulPPelaez (approved):
This looks great to me. I say lets merge.
BTW, I see this is suffering from #98 
Do you have any explanation?
I have not been able to reproduce it. My guess is on some CMake parameter.

raimis (approved):
Looks good!

PR #103, opened 2023-05-23
Particle Mesh Ewald

peastman:
I'm working on an implementation of PME.  This will be useful for models that include explicit Coulomb terms, especially when the charges are predicted by the model and require extra chain rule terms.

The direct space interaction is implemented in PyTorch and is very simple.  The reciprocal space part is implemented as a C++ extension.  I've written the CPU version.  I still have to do the CUDA version.


proteneer:
Would it be possible to implement the reciprocal space energy calculations in python/torch as well? It would be easier to do tests of things like forces by taking advantage of automatic differentiation (as opposed to just comparing against hard coded floats)

peastman:
This is ready for review.

I originally hoped to implement the direct space calculation in Python, but I couldn't figure out any way to do it that was compatible with TorchScript.  The implementation still isn't especially fast, since it just uses the NNPOps neighbor list structure, which isn't well suited to GPU computation, but it should be plenty fast enough for our purposes.  The reciprocal space calculation is based on the implementation in OpenMM.  I simplified it a bit, removing optimizations that add a lot of complexity for only modest speedup, but the performance should still be pretty good.

If someone can double check my values for Coulomb's constant, that would be helpful!

src/pytorch/pme/pme.py
```diff
@@ -0,0 +1,194 @@
...
+                arg = (2*torch.pi*i/ndata)*torch.arange(ndata)
+                sc = torch.sum(bsplines_data[:ndata]*torch.cos(arg))
+                ss = torch.sum(bsplines_data[:ndata]*torch.sin(arg))
+                m[i] = sc*sc + ss*ss
+            for i in range(ndata):
+                if m[i] < 1e-7:
+                    m[i] = (m[(i-1+ndata)%ndata]+m[(i+1)%ndata])*0.5
+            self.moduli.append(m)
+
+    def compute_direct(self, positions: torch.Tensor, charges: torch.Tensor, cutoff: float, box_vectors: torch.Tensor, max_num_pairs: int = -1):
```
RaulPPelaez, line 129: The grid size is specified in the constructor, I guess for performance reasons. Does it really make sense then to leave the box size change at every forward call? grid size and box size are kind of tied to each other, right?

src/pytorch/pme/pme.py
```diff
@@ -0,0 +1,194 @@
...
+            raise ValueError('positions, charges, and exclusions must all have the same length')
+        if box_vectors.dim() != 2 or box_vectors.shape[0] != 3 or box_vectors.shape[1] != 3:
+            raise ValueError('box_vectors must have shape (3, 3)')
+        if (cutoff <= 0):
+            raise ValueError('cutoff must be positive')
+        neighbors, deltas, distances, number_found_pairs = getNeighborPairs(positions, cutoff, max_num_pairs, box_vectors)
+        self.exclusions = self.exclusions.to(positions.device)
+        return torch.ops.pme.pme_direct(positions, charges, neighbors, deltas, distances, self.exclusions, self.alpha, self.coulomb)
+
+    def compute_reciprocal(self, positions: torch.Tensor, charges: torch.Tensor, box_vectors: torch.Tensor):
```
RaulPPelaez, line 165: Besides testing purposes, is there any situation in which one would call compute_direct() and compute_reciprocal() for any other thing apart from subtracting each other? 
I believe it would be useful for the PmeModule you wrote for the test to be defined in this file instead.

src/pytorch/pme/pmeCUDA.cu
```diff
@@ -0,0 +1,447 @@
+#include <torch/extension.h>
+#include <c10/cuda/CUDAGuard.h>
+#include <c10/cuda/CUDAStream.h>
+#include <array>
+#include <cmath>
+#include <vector>
+
+#include "common/accessor.cuh"
```
RaulPPelaez, line 8: This file is not part of the PR right now.

src/pytorch/pme/pmeCUDA.cu
```diff
@@ -0,0 +1,447 @@
...
+    // Subtract excluded interactions to compensate for the part that is
+    // incorrectly added in reciprocal space.
+
+    float dr[3];
+    for (int index = blockIdx.x*blockDim.x+threadIdx.x; index < numAtoms*maxExclusions; index += blockDim.x*gridDim.x) {
+        int atom1 = index/maxExclusions;
+        int atom2 = exclusions[atom1][index-atom1*maxExclusions];
+        if (atom2 > atom1) {
+            for (int j = 0; j < 3; j++)
+                dr[j] = pos[atom1][j]-pos[atom2][j];
```
RaulPPelaez, line 74: Dont you have to apply PBC here?

src/pytorch/pme/pmeCUDA.cu
```diff
@@ -0,0 +1,447 @@
...
+        data[0][i] = scale*(1-dr[i])*data[0][i];
+    }
+}
+
+template <int PME_ORDER>
+__global__ void spreadCharge(const Accessor<float, 2> pos, const Accessor<float, 1> charge, const Accessor<float, 2> box,
+                             Accessor<float, 3> grid, int gridx, int gridy, int gridz, float sqrtCoulomb) {
+    float recipBoxVectors[3][3];
+    invertBoxVectors(box, recipBoxVectors);
+    float data[PME_ORDER][3];
```
RaulPPelaez, line 163: The number of registers in these kernels is really high. I wonder if computing these in another kernel and fetching from global memory here would be better.

src/pytorch/pme/pmeCUDA.cu
```diff
@@ -0,0 +1,447 @@
...
+        for (int j = 1; j < pmeOrder-1; j++)
+            data[pmeOrder-j-1][i] = scale*((dr[i]+j)*data[pmeOrder-j-2][i]+(pmeOrder-j-dr[i])*data[pmeOrder-j-1][i]);
+        data[0][i] = scale*(1-dr[i])*data[0][i];
+    }
+}
+
+template <int PME_ORDER>
+__global__ void spreadCharge(const Accessor<float, 2> pos, const Accessor<float, 1> charge, const Accessor<float, 2> box,
+                             Accessor<float, 3> grid, int gridx, int gridy, int gridz, float sqrtCoulomb) {
+    float recipBoxVectors[3][3];
```
RaulPPelaez, line 161: You would save 9 registers per thread by placing this in shared memory and letting the first thread compute or simply by passing it as an argument.

src/pytorch/pme/pmeCUDA.cu
```diff
@@ -0,0 +1,447 @@
...
+                atomicAdd(&posDeriv[atom2][j], -dEdR*dr[j]);
+            }
+        }
+    }
+    energyBuffer[blockIdx.x*blockDim.x+threadIdx.x] = energy;
+}
+
+__device__ void invertBoxVectors(const Accessor<float, 2>& box, float recipBoxVectors[3][3]) {
+    float determinant = box[0][0]*box[1][1]*box[2][2];
+    float scale = 1.0/determinant;
```
RaulPPelaez, line 99: 1.0f so it does not get promoted to double and then converted back to float

src/pytorch/pme/pmeCUDA.cu
```diff
@@ -0,0 +1,447 @@
...
+        int gridSize[3] = {(int) gridx.toInt(), (int) gridy.toInt(), (int) gridz.toInt()};
+        float sqrtCoulomb = sqrt(coulomb.toDouble());
+
+        // Spread the charge on the grid.
+
+        TensorOptions options = torch::TensorOptions().device(positions.device());
+        Tensor realGrid = torch::zeros({gridSize[0], gridSize[1], gridSize[2]}, options);
+        int blockSize = 128;
+        int numBlocks = max(1, min(getMaxBlocks(), (numAtoms+blockSize-1)/blockSize));
+        TORCH_CHECK(pmeOrder == 4 || pmeOrder == 5, "Only pmeOrder 4 or 5 is supported with CUDA");
```
RaulPPelaez, line 343: Why this restriction?

src/pytorch/pme/pmeCUDA.cu
```diff
@@ -0,0 +1,447 @@
...
+        // Store data for later use.
+
+        ctx->save_for_backward({positions, charges, box_vectors, xmoduli, ymoduli, zmoduli, recipGrid});
+        ctx->saved_data["gridx"] = gridx;
+        ctx->saved_data["gridy"] = gridy;
+        ctx->saved_data["gridz"] = gridz;
+        ctx->saved_data["order"] = order;
+        ctx->saved_data["alpha"] = alpha;
+        ctx->saved_data["coulomb"] = coulomb;
+        return {0.5*torch::sum(energy)};
```
RaulPPelaez, line 371: You would save a kernel launch by putting the 0.5 in reciprocalConvolution, maybe call the tensor half_energy then.

src/pytorch/pme/pmeCPU.cpp
```diff
@@ -0,0 +1,383 @@
...
+        int numAtoms = positions.size(0);
+        auto box = box_vectors.accessor<float,2>();
+        auto pos = positions.accessor<float,2>();
+        auto charge = charges.accessor<float,1>();
+        float recipBoxVectors[3][3];
+        invertBoxVectors(box_vectors, recipBoxVectors);
+
+        // Take the inverse Fourier transform.
+
+        Tensor realGrid = torch::fft::irfftn(recipGrid, {gridSize[0], gridSize[1], gridSize[2]}, c10::nullopt, "forward");
```
RaulPPelaez, line 302: torch<2 refuses to compile this line because it cannot convert from ```{gridSize[0], gridSize[1], gridSize[2]}``` to the expected ```c10::optional<c10::ArrayRef<long int> >``` type. 
Maybe passing ``` int gridSizeTemp[3] = {gridSize[0], gridSize[1], gridSize[2]}``` makes it happy?

RaulPPelaez:
Really clean implementation @peastman ! Just some comments/questions from me, but looks great!

src/pytorch/pme/pme.py
peastman: One common situation is multiple time step integrators, where you evaluate the direct space part more often than the reciprocal space part.  I also want to experiment with models where I include only the reciprocal space part, which covers long range interactions for which point charges are a pretty good approximation, while letting the neural network try to learn a more accurate short range part.

src/pytorch/pme/pmeCUDA.cu
peastman: Exclusions apply only to a single copy of each other atom, the one in the same periodic copy.  See the discussion of this in the class documentation.

src/pytorch/pme/pmeCUDA.cu
peastman: I see you had the same idea. :)

src/pytorch/pme/pmeCUDA.cu
peastman: The order is a template parameter to the kernel, so we have to specify all supported values at compile time.  4 and 5 are the only commonly used orders.

RaulPPelaez: Oh right, it is not a template CPU side.
If you think its worth, I usually dispatch things like that with something similar to this:
```c++
auto spread_kernel = spreadCharge<1>;
switch(pmeOrder){
       case 2: spread_kernel = spreadCharge<2>;
       case 3: spread_kernel = spreadCharge<3>;
       case 4: spread_kernel = spreadCharge<4>;
       ...
}
spread_kernel<<<...>>>(/*parameters*/)
```
Avoids having to copy all arguments.

peastman: Good idea, except you're missing the `break` statements.  (I tend to avoid `switch` in C++ because it's so easy to make that mistake.)

src/pytorch/pme/pmeCPU.cpp
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    static tensor_list backward(AutogradContext *ctx, tensor_list grad_outputs) {
+        auto saved = ctx->get_saved_variables();
+        Tensor posDeriv = saved[0];
+        Tensor chargeDeriv = saved[1];
+        torch::Tensor ignore;
+        return {posDeriv*grad_outputs[0], chargeDeriv*grad_outputs[0], ignore, ignore, ignore, ignore, ignore, ignore};
+    }
+};
```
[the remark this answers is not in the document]
peastman: I'm not sure it's possible to support second derivatives, except by writing code to compute them explicitly.  This corresponds to the last example in the page you linked.  Since the calculations are done by custom code, PyTorch doesn't know how to define gradients.

src/pytorch/pme/pmeCUDA.cu
peastman: Done.

src/pytorch/pme/pmeCUDA.cu
peastman: Done.

src/pytorch/pme/pmeCUDA.cu
peastman: Done.

src/pytorch/pme/pmeCUDA.cu
peastman: Done.

src/pytorch/pme/pmeCPU.cpp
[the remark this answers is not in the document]
RaulPPelaez: This backwards is quite involved, the only way I see is to write the double backwards yourself, as you say.
The dangerous thing for me is that it will be silently wrong if you try to grad twice. Maybe there is a short way to at least raise an error if something tries to double backwards. Sadly, as far as I can see, even in the most simple form you need to implement a second autograd function like PME_Backwards.

src/pytorch/pme/pme.py
```diff
@@ -0,0 +1,194 @@
...
+    used for nearby atoms within the same molecule.  When two atoms are listed as an exclusion, only the interaction of
+    each with the same periodic copy of the other (that is, not applying periodic boundary conditions) is excluded.
+    Each atom still interacts with all the periodic copies of the other.
+
+    Due to the way the reciprocal space term is calculated, it is impossible to prevent it from including excluded
+    interactions.  The direct space term therefore compensates for it, subtracting off the energy that was incorrectly
+    included in reciprocal space.  The sum of the two terms thus yields the correct energy with the interaction fully
+    excluded.
+
+    When performing backpropagation, this class computes derivatives with respect to atomic positions and charges, but
```
RaulPPelaez, line 35: Maybe I am over amplifying this because it made me waste a lot of time recently, but I believe there should be a mention here to this module having only a single backwards pass.
If one trains with forces this will be grad'ed during the forward pass, which requires second grad during backprop.

peastman: Yeah, that's an annoying limitation.  It shouldn't be an issue for the sorts of problems I'm interested in, since you wouldn't use PME during training.  You would be training on isolated (not periodic) small molecules, for which all interactions are computed in direct space.  PME would only be used for inference, if you want to apply the model to larger systems.  It would be more of an issues for materials applications, if you wanted to train on crystals.

I'll see if I can get it to throw an exception if you try to take a second derivative.

RaulPPelaez: Python-side there is the @once_differentiable decorator: https://discuss.pytorch.org/t/what-does-the-function-wrapper-once-differentiable-do/31513

peastman: It turns out that it already throws an exception:

```
RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

I added a test case to check for it.

PR #105, opened 2023-06-27
Install via pip

RaulPPelaez:
This PR is an effort to include the possibility of installing NNPops via pip.

I mainly took the example here and adapted it:
https://github.com/pybind/cmake_example/blob/master/setup.py

Having pip as an option would simplify installing NNPops in environments which are not covered by the conda feedstock.
If this is not something that fits the current view of the project, please feel free to close this PR, @peastman.
 
I added a "pip wheel" section to the CI, lets see if it runs...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,3 +129,13 @@ jobs:
         conda activate nnpops
         cd build
         ctest --verbose --exclude-regex TestCuda
+        
+    - name: Pip build (without CUDA)
+      if: ${{ !matrix.enable_cuda }}
+      run: |
+        pip -vvv wheel --wheel-dir=$(pwd) nnpops --extra-index-url https://download.pytorch.org/whl/cpu
+    - name: Pip build (with CUDA)
+      if: ${{ matrix.enable_cuda }}
+      run: |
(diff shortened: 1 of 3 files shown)


peastman:
Being able to install with pip would be great.  Eventually we want to make the whole OpenMM stack pip installable.

RaulPPelaez:
I can do this from my laptop:
```shell
$ mamba create -n nnpops && mamba activate nnpops
$ mamba install cudatoolkit-dev==11.7
$ pip install git+https://github.com/openmm/NNPOps@refs/pull/105/head
```
There are however some things I still do not know how to solve:
1. A valid CUDA and gcc must be present in the system. AFAIK you cannot install that via pip, nor modify pyproject.toml depending on the environment.
2. Link CUDA with dependencies.
   Currently pip will install torch>=2 to build regardless of the CUDA version in the system, which will easily result in a failed build. Ideally pip should decide to install a lower version of torch depending on the available CUDA version.

peastman:
The standard way of building pip packages is with [cibuildwheel](https://github.com/pypa/cibuildwheel).  It provides a set of standard build environments to ensure compatibility between packages.

CUDA can be [installed with pip](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#pip-wheels).

PR #115, opened 2023-10-14
Add OSX Support

SimonBoothroyd:
This PR updates how the custom ops lib is loaded to support OSX (and in theory Windows), and enables OSX tests in the CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,46 +23,46 @@ jobs:
         include:
 
           # Oldest supported versions
-#          - name: Linux (CUDA 10.2, Python 3.8, PyTorch 1.11)
-#            os: ubuntu-18.04
-#            enable_cuda: true
-#            cuda: "10.2.89"
-#            gcc: "8.5.*"
-#            nvcc: "10.2"
-#            python: "3.8.*"
-#            torchani: "2.2.*"
-#            pytorch: "1.11.*"
-#
-#          # Older supported versions
-#          - name: Linux (CUDA 11.2, Python 3.9, PyTorch 1.12)
-#            os: ubuntu-18.04
-#            enable_cuda: true
-#            cuda: "11.2.2"
-#            gcc: "10.3.*"
-#            nvcc: "11.2"
-#            python: "3.9.*"
-#            torchani: "2.2.*"
-#            pytorch: "1.12.*"
-#
-#          # Latest supported versions (with CUDA)
-#          - name: Linux (CUDA 11.8, Python 3.10, PyTorch 2.0)
-#            os: ubuntu-18.04
-#            enable_cuda: true
-#            cuda: "11.8.0"
-#            gcc: "10.3.*"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -15,15 +15,16 @@ defaults:
...
     name: ${{ matrix.name }}
-    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
         include:
 
           # Oldest supported versions
           - name: Linux (CUDA 10.2, Python 3.8, PyTorch 1.11)
+            os: ubuntu-18.04
```
RaulPPelaez, line 27: Why go back to ubuntu-18 from 22?

SimonBoothroyd: 🤦 good catch, thanks! fixed in 5b4051a

.github/workflows/ci.yml
```diff
@@ -100,6 +111,16 @@ jobs:
                -e "/pytorch-gpu/c\  - pytorch-cpu ${{ matrix.pytorch }}" \
                environment.yml
 
+    - name: Prepare dependencies (Mac)
+      if: ${{ contains(matrix.os, 'macos') }}
+      run: |
+        sed -i '' -e "s/- cudatoolkit/# - cudatoolkit/" \
```
RaulPPelaez, line 117: It really looks like you are trying to install cudatoolkit here in OSX.
Not sure how to make it more obvious that you are commenting out the line.
Perhaps `sed -i '' "/- cudatoolkit/d"`

SimonBoothroyd: hmm yeah I see what you mean! this should be swapped over in 4014ff4

.github/workflows/ci.yml
```diff
@@ -100,6 +111,17 @@ jobs:
                -e "/pytorch-gpu/c\  - pytorch-cpu ${{ matrix.pytorch }}" \
                environment.yml
 
+    - name: Prepare dependencies (Mac)
+      if: ${{ contains(matrix.os, 'macos') }}
+      run: |
+        sed -i '' -e "/- cudatoolkit/d" \
+                  -e "/- gxx_linux-64/d" \
```
RaulPPelaez, line 118: You dont need to install some version of gxx for Mac?

RaulPPelaez:
@peastman, I believe someone has to manually allow the new CI to run. Could you do so?

peastman:
It already ran without me doing anything.  Two builds failed, one from CUDA not being available and one from running out of disk space.

SimonBoothroyd:
> However the 10.2 one is strange. CUDA is not available but that should not be a problem. In fact it seems like the Doctest is interpreting the warning as an error.

Hmm strange... can someone re-trigger the CI to see if this failure was a one off due to some runner issue?

> currently the CI is picking the C++ compiler from the docker runner.

Sure I can swap to conda distributed compilers if you'd prefer - should be swapped in 1794c6e

PR #118, opened 2024-07-08
Use args when calling self.aev_computer

lohedges:
This PR allows the use of a forward hook on the internal AEVComputer of ANI models when using OptmizedTorchANI via TorchScript. See aiqm/torchani#649 for details.

### DIFF
--- a/src/pytorch/OptimizedTorchANI.py
+++ b/src/pytorch/OptimizedTorchANI.py
@@ -49,7 +49,7 @@ class OptimizedTorchANI(torch.nn.Module):
                 pbc: Optional[Tensor] = None) -> SpeciesEnergies:
 
         species_coordinates = self.species_converter(species_coordinates)
-        species_aevs = self.aev_computer(species_coordinates, cell=cell, pbc=pbc)
+        species_aevs = self.aev_computer(species_coordinates, cell, pbc)
         species_energies = self.neural_networks(species_aevs)
         species_energies = self.energy_shifter(species_energies)
 


peastman:
Does this introduce any compatibility issues we need to be aware of, such as only working with newer versions of TorchANI?

lohedges:
I don't think so, I've checked back to version 1.0.0  of TorchANI and the signature of `AEVComputer.forward` is identical (although missing type hinting). Older versions, e.g. the original 0.1 release, are missing the `cell` and `pbc` kwargs, but I doubt you support those given that the oldest version of TorchANI on conda-forge is 2.2.

Just to note that while the PR linked to above won't (yet) be merged due to a code freeze, this fix is independent of it, i.e. it allows you to use an AEVComputer forward hook with OptimizedTorchANI without requiring the fix to TorchANI itself.

RaulPPelaez (approved):

peastman:
@RaulPPelaez any idea what's going on with CI?

peastman:
I checked and the rules listed there are correct.  The problem is that none of the CI builds have run.  And there's no option to retrigger them.  Let me try making a trivial change to the file and see if that gets it working.

peastman:
Look like they're running now.

peastman:
One build is failing with the dreaded `[Errno 28] No space left on device'`.  It's clearly not related to this change, so I'm ok with merging it.  I'm currently dealing with that same error in [trying to do a new OpenMM dev build](https://github.com/conda-forge/openmm-feedstock/pull/132).

lohedges:
Many thanks for looking at this so quickly.

PR #122, opened 2025-01-30
Update CI versions

peastman:
We were running CI against very old versions of Python, CUDA, and PyTorch.  This updates it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,26 +23,6 @@ jobs:
         include:
 
           # Oldest supported versions
-          - name: Linux (CUDA 10.2, Python 3.8, PyTorch 1.11)
-            enable_cuda: true
-            cuda: "10.2.89"
-            gcc: "8.5.*"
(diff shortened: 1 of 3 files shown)


peastman:
Both of the CUDA builds run out of disk space trying to install everything.  I copied over the code we use with OpenMM-Torch to try to free up space, but it isn't enough.

Only the no-CUDA build manages to complete, but it gets a lot of test failures.  They all seem to be problems with compiling TorchANI, variations on

> RuntimeError: Could not cast attribute '_species_to_tensor' to type \_\_torch\_\_.torchani.utils.ChemicalSymbolsToInts (of Python compilation unit at: 0x55b61d384a30): Could not cast attribute 'rev_species' to type Dict[Tensor, int]: Unable to cast H to Tensor

peastman:
I have it to the point where the CUDA 11 build passes, and the CUDA 12 build successfully compiles, but tests fail when trying to compile TorchANI.  I'm going to merge this and debug that problem in a separate PR.

PR #124, opened 2025-02-12
Make NNPOps pip-installable

stefdoerr:
Update NNPOps to compile with setuptools instead of CMake and reorganize it to install correctly as a pip package


peastman:
Why did you completely reorganize the whole repository?  That isn't needed just to build with pip.

stefdoerr:
It's true that it's not strictly necessary but I didn't modify any code files, only changed the folder structure.
The reason was that the codebase was a bit of a mess, there are test files inside `src` directory and even executable benchmark files inside an importable python module which means that if you for example press Tab in IPython for autocompletion it could start running the benchmarks.
So I decided to choose the structure which is easiest for creating a python package with setuptools without picking and choosing each file which ends in the package. If you disagree with it I can try to make it work without that.

image

stefdoerr:
Practically, since it's not visible in the diff, I went for this folder structure where benchmarks, tests and molecules are outside of the source-code folder. 
image

PR #125, opened 2025-02-17
Pip package 2

stefdoerr:
Here is a new PR for making the repo pip installable. I tried to not rename any folders but that interferes with the automatic discovery of setuptools and I got tired of trying to make it rename the module from `pytorch` to `NNPOps`.
https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#automatic-discovery

I tend to go for flat-layout since it's easily importable from a REPL (as in my last PR)
https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#flat-layout

But I ended up going with the src-layout in this PR since it is a bit closer to the layout here
https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#src-layout

But this required that I rename the pytorch folder to NNPOps which is the only major change here.


stefdoerr:
In the previous PR I had also modified the CI so that it works with setuptools instead of CMake so that it's tested accordingly but I leave it up to you if you want that change.

peastman:
Same as I said on #124: please don't just create a PR reorganizing everything.  Open an issue, propose an organization, and we can discuss it first.

stefdoerr:
I guess I don't consider renaming a single folder to maintain a setuptools standard as reorganizing everything. But ok I found a way to make it work by moving the setup files into the src directory and renaming the module.

stefdoerr:
@peastman can you take another look now?

peastman:
Tests are failing with this error:

> 8: FAILED test/TestSpeciesConverter.py::test_model_serialization[1hvj-cpu] - RuntimeError: Could not cast attribute '_species_to_tensor' to type __torch__.torchani.utils.ChemicalSymbolsToInts (of Python compilation unit at: 0x5578a9270330): Could not cast attribute 'rev_species' to type Dict[Tensor, int]: Unable to cast H to Tensor

This generally means it's installing the wrong version of torchani.  See https://github.com/openmm/openmm-ml/issues/23.

src/pyproject.toml
```diff
@@ -0,0 +1,30 @@
+[project]
+name = "NNPOps"
+description = "Highly optimized, open source implementations of bottleneck operations in NNPs"
+authors = [{ name = "OpenMM", email = "info@openmm.org" }]
```
peastman, line 4: I don't think that address exists?

stefdoerr: Yes you are right, it was just a placeholder since I didn't know what to put. Feel free to suggest an email

stefdoerr:
> This generally means it's installing the wrong version of torchani. See https://github.com/openmm/openmm-ml/issues/23.

Hm, why? We are not installing the pip package anywhere. I didn't modify the CI.

stefdoerr:
The main branch of the repo is failing with the same error. It's not related to this PR as far as I can tell.

src/pyproject.toml
peastman: Let's not list an email.  The way for people to contact us is through the github repo.

stefdoerr: Done

PR #126, opened 2025-12-05
Fixed errors installing with pip

peastman:
This fixes two errors from #125.  First, it hardcoded a specific version of PyTorch.  Second, it processed CUDA architecture versions incorrectly.  I also fixed a bug that caused `import NNPOps` to fail if torchani wasn't installed.

### DIFF
--- a/src/pyproject.toml
+++ b/src/pyproject.toml
@@ -9,7 +9,7 @@ classifiers = [
   "Programming Language :: Python :: 3",
   "Operating System :: POSIX :: Linux",
 ]
-dependencies = ["torch==2.5.1.*"]
+dependencies = ["torch"]
 
 
 [project.urls]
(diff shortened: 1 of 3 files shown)


epretti:
This didn't fix the problem of being able to `import NNPOps` without torchani for me.  I also had to remove torchani imports from `BatchedNN.py`, `EnergyShifter.py`, `SpeciesConverter.py`, and `SymmetryFunctions.py`.

It appears that these are only used for type checking?  If we wanted to keep these type annotations around, I think one possibility is to use `__future__.annotations` and conditionally import torchani based on `typing.TYPE_CHECKING`.  I don't know how much value there is in keeping them, though.

peastman:
Go ahead and remove them.  They should never have been there in the first place.  See #39 for background.

